PR #28, opened 2018-05-01
Fix build

shawwn:


modulesio:
Hm, does this actually do anything? It seems like a no-op.

It might be something to do with `nvm-windows` `binding.gyp` environment variables though...

shawwn:
Nope. Fixed by 75b3d94e27b537384f661baa5014b0ac8d513e67

PR #29, opened 2018-05-03
Add -v argument

shawwn:


index.js
```diff
@@ -62,15 +64,22 @@ const args = (() => {
...
 })();
+
 core.setArgs(args);
-core.setVersion(version);
+const agent = core.setVersion(version);
 core.setNativeBindingsModule(nativeBindingsModulePath);
 
+if (args.version) {
+  console.log(agent);
+  process.exit(0);
```
modulesio: Can we make this decision earlier? Seems a lot of native-ish work is happening above even though we won't be using it.

shawwn: depends whether core.js or index.js should be responsible for the name: 

https://github.com/webmixedreality/exokit/pull/29/files#diff-04aff9558c8839f145fd5cc256639700R42

```let agent = {browser: 'MixedReality', name: 'Exokit', version: '0.0.0'}```

shawwn: browser/name should be specified in package.json anyway, next to version.

modulesio: I disagree with browser. It’s not a ptoperty of the module. I would hardcode that string.

shawwn: How about we just convert the package.json name to CamelCase then and use that as name? Then there's nothing left to specify.

modulesio: package.json should have nothing to do with functionality.

Also I don’t think that has anything to do with the review concern...

shawwn: I was thinking core.js should take in a user agent string as a parameter, and `--version` should print it and exit.

If it's parameterized then it can happen at the top of the file.

Then I realized it's a PITA to parse a user agent string compared to JSON, so I printed it out directly. Now I'm not sure which is better.

core.js builds the user agent string right now. If it stays in charge of that, then we can't return all necessary information in `--version` until after core.js runs.

parameterizing the useragent string seems best, given exokit's goals.

modulesio: The version isn’t the ua string to me. Only the package.json version.

shawwn:
See #35.

PR #36, opened 2018-05-09
Remove textContent setter from HTMLScriptElement

shawwn:


core.js
```diff
@@ -2174,6 +2174,10 @@ class HTMLScriptElement extends HTMLLoadableElement {
     this.innerHTML = textContent;
   }
 
+  get textContent() {
```
modulesio: I don't think this is what the spec says. If there is a missing accessor it should probable be on some parent class, not Script.

core.js
```diff
@@ -2152,28 +2152,6 @@ class HTMLScriptElement extends HTMLLoadableElement {
     this.setAttribute('type', type);
   }
 
-  get text() {
```
modulesio: Ok, have we checked that this continues to behave per spec when you mutate the `<script/>` and such?

This seems like something that would break a _lot_ of things.

core.js
```diff
@@ -2170,10 +2170,6 @@ class HTMLScriptElement extends HTMLLoadableElement {
     this._emit('innerHTML', innerHTML);
   }
 
-  set textContent(textContent) {
```
modulesio: I still don't think this is right, since setting the src probably won't run the script now. Should it?

What does the spec say?

shawwn: https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement

image

Chrome doesn't re-execute scripts that mutate via .textContent, so I think this PR is ready to merge.

PR #39, opened 2018-05-09
Use > as the REPL prompt for compatibility

shawwn:


shawwn:
One reason to do this is because IDEs are set up to easily integrate with REPLs that use > as the prompt.

(E.g. emacs)

modulesio:
Interesting... can we flag it?

shawwn:
Let's just use > as the prompt. No need to add another flag

modulesio:
I kind of like the flair of having it as is :P.

I don't think emacs should be dictating the prompt format; there is not really a prompt standard to follow here.

shawwn:
Up to you, but my point was that a lot of code has been written for the case of "Spawn a child process and wait for `> `"

Not just emacs, but pretty much any code that controls a REPL interactively via child process.

shawwn:
Emojis win the day.

PR #41, opened 2018-05-09
Add --interactive argument

shawwn:


index.js
```diff
@@ -1131,16 +1141,8 @@ const _start = () => {
...
-    let window = null;
-    const _bindReplWindow = newWindow => {
-      _bindWindow(newWindow, _bindReplWindow);
-      window = newWindow;
-    };
-    _bindReplWindow(core('', {
-      dataPath,
-    }));
-
+  }
```
modulesio: This is pretty weird.

index.js
```diff
@@ -1056,6 +1059,15 @@ const _prepare = () => Promise.all([
   }),
 ]);
 const _start = () => {
+  let window = null;
```
modulesio: Why does this need to be here at all? Seems the `.interactive`  already does what this PR advertises. :/

shawwn: because `node .`

modulesio: But then you will have `!url`.

What I'm saying is therefore it seems that the only thing we need is [this line](https://github.com/webmixedreality/exokit/pull/41/files#diff-168726dbe96b3ce427e7fedce31bb0bcR1148) plus the args declarations.

modulesio: I don't think this `window` latching here is right. Doing this without a full `_bindReplWindow` sounds like it would be wrong and introduce navigation bugs.

Has this been tested when changing the URL programmatically via mechanisms like `.href =` on the user site?

shawwn:
Closed as unimportant.

PR #56, opened 2018-05-10
Let exokit-macos take arguments

shawwn:


exokit-macos
```diff
@@ -14,4 +14,4 @@ if [ -x "$DIR/node/bin/node" ]; then
 fi
 
 cd "$DIR"
-exec "$NODE_BIN" "$DIR/index.js" -h
```
modulesio: Couldn't this just be `-h "$@"`? That's what the other places use I think.

shawwn: No, that would always pass in -h.

shawwn:
It’s probably a better idea to refactor this into exokit.sh and use that instead, since it’s identical code.

modulesio:
Hm, yeah; the only reason it was done in this script instead is because of MacOS `.app` requirements for what's considered a "binary" -- and it can't be called `.sh`.

I'll patch this into both scripts. Thanks!

PR #58, opened 2018-05-10
Codesign Exokit.app

### LINKED ISSUE #55: MacOS .app bundle is unsigned
Downloading from get.webmr.io yields the following error:

image

### LINKED ISSUE #57: CI: do not codesign PRs
Currently our builds will error during a PR build: https://travis-ci.org/webmixedreality/exokit/builds/377086347?utm_source=slack&utm_medium=notification

This is by design: the PRing Github account does not (and should not) have access to secure variables in the build, such as our code signing keys. But it would be nice for the build to not fail, since build success is highly useful info for PRs.

The straightforward fix for this would be to skip the code signing steps (for [Windows on Appveyor](https://github.com/webmixedreality/exokit/blob/master/appveyor.yml#L39) and [OSX on Travis CI](https://github.com/webmixedreality/exokit/blob/master/.travis.yml#L74) if/when the `PFX_KEY` environment variable is not defined -- and maybe log the fact that this is happening for later debugging in case this behavior leaks into `master`.

CC @shawwn

shawwn:
Closes #55, closes #57.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,7 @@ matrix:
     - tar -czf exokit-macos-full.tar.gz --exclude=".*" --exclude="*.tar.gz" *
     - rm -R node
     - tar -czf exokit-macos.tar.gz --exclude=".*" --exclude="*.tar.gz" *
-    - sh codesign-macos.sh
+    - ./codesign-macos.sh
     deploy:
     - provider: releases
       api_key:
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,18 +63,10 @@ matrix:
     - mkdir -p /tmp/Exokit.app/Contents/MacOS
     - cp -R * /tmp/Exokit.app/Contents/MacOS
     - cp Info.plist /tmp/Exokit.app/Contents
-    - cp appdmg.json /tmp
     - tar -czf exokit-macos-full.tar.gz --exclude=".*" --exclude="*.tar.gz" *
     - rm -R node
     - tar -czf exokit-macos.tar.gz --exclude=".*" --exclude="*.tar.gz" *
-    - security create-keychain -p nopassword build.keychain
-    - security default-keychain -s build.keychain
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -63,17 +63,10 @@ matrix:
...
     - rm -R node
     - tar -czf exokit-macos.tar.gz --exclude=".*" --exclude="*.tar.gz" *
-    - security create-keychain -p nopassword build.keychain
-    - security default-keychain -s build.keychain
-    - security unlock-keychain -p nopassword build.keychain
-    - security set-keychain-settings -t 3600 -u build.keychain
-    - security import codesign-macos.pfx -k build.keychain -P "$PFX_KEY" -A
-    - "security set-key-partition-list -S apple-tool:,apple: -s -k nopassword build.keychain"
-    - appdmg /tmp/appdmg.json exokit-macos-x64.dmg
+    - sh codesign-macos.sh
```
modulesio: Can we `./`?

shawwn: I'd prefer not to rely on chmod +x for `codesign-macos.sh`, but it's up to you.

shawwn: Okay, done.

shawwn:
set -x would leak PFX_KEY into the build logs.

modulesio:
@shawwn as far as CI goes, it would be filtered from logs. It’s an automatic Travis feature afair.

Additionally, would it not be possible to devnull it even if that weren’t the case?

shawwn:
My main issue with this setup is that it takes a full hour to sign all the files in .app, and then doesn’t successfully sign it anyway. It complains about node_modules/.bin

The PR might be fine as a way of getting rid of build errors for contributors, but perhaps not super important.

A better approach is probably to zip node_modules somehow. See Zeit/pkg

modulesio:
> A better approach is probably to zip node_modules somehow. See Zeit/pkg

Hm, interesting approach...

PR #60, opened 2018-05-12
WebXR Support

modulesio:
#MozSprint


modulesio:
Closes https://github.com/webmixedreality/exokit/issues/51

PR #61, opened 2018-05-18
Add initial XMLSerializer implementation

shawwn:
This PR uses [xmlserializer](https://www.npmjs.com/package/xmlserializer) to implement the [XMLSerializer API](https://developer.mozilla.org/en-US/docs/Web/API/XMLSerializer).


core.js
```diff
@@ -1633,6 +1633,12 @@ const _defineId = (window, id, el) => {
     configurable: true,
   });
 };
+const _XMLSerializer = require("xmlserializer");
+class XMLSerializer {
+  serializeToString(dom) {
+    return _XMLSerializer.serializeToString(parse5.parse(dom.innerHTML));
```
modulesio: I don't think this is right. I realize `xmlserializer` wants a `parse5` tree but if you do `dom.innerHTML` and/or `parse5.parse`, that seems like it would go through `tagName` normalization and other HTML-isms.

This would probably also be very slow for large XML's, as well difficult to debug, since it turns a serialization into a serialization plus a parse plus another serialization...

I think the correct solution is to massage the nodes into a parse5 tree, or use an adapter instead.

PR #62, opened 2018-05-22
Fix stylesheet parsing

shawwn:
This fixes https://threejs.org/examples/webgl_loader_svg.html


modulesio:
Since PR doesn't mention this:

The reason this works is that the SVG rendering was using style attributes, which were not correctly populated in the style proxy. This fixes that by doing an initial lazy population when a style attribute is accessed.

PR #64, opened 2018-05-23
Make HTMLCanvasElement return numeric width and height

shawwn:
This fixes https://threejs.org/examples/webgl_materials_texture_canvas.html


modulesio:
Thanks 👏

PR #66, opened 2018-05-23
[WIP] Silence mlsdk lifecycle spam

shawwn:
Could you try this out and see if it works for you?

This PR silences the periodic `"Lifecycle library is not connected."` message.


deps/exokit-bindings/magicleap/src/magicleap.cc
```diff
@@ -630,7 +739,52 @@ void MLContext::LifecycleInit() {
   lifecycle_callbacks.on_pause = onPause;
   lifecycle_callbacks.on_resume = onResume;
 
-  lifecycle_status = MLLifecycleInit(&lifecycle_callbacks, (void*)&application_context);
+#if 1
+  {
+    std::ofstream filestr;
+    std::ofstream errstr;
+    //std::cout << "Redirecting stdout\n";
+    std::streambuf* oldsb = redirect_output("mlsdk_log1.txt", filestr, std::cout);
```
modulesio: Any way to not force a file write?

deps/exokit-bindings/magicleap/src/magicleap.cc
```diff
@@ -622,6 +626,111 @@ NAN_METHOD(MLContext::OnPresentChange) {
...
+
+#include <stdarg.h>
+#include <fcntl.h>
+#include <stdio.h>
+#if _MSC_VER
+#include <io.h>
+#define STDOUT_FILENO 1
+#define STDERR_FILENO 2
+#else
+#include <unistd.h>
```
modulesio: Was this checked on OSX?

shawwn: Not yet.

deps/exokit-bindings/magicleap/src/magicleap.cc
shawwn: cout and cerr can be redirected to std::ostringstream, but I'm not sure about stdout and stderr. It might require file handles.

modulesio:
Closing as outdated.

PR #67, opened 2018-05-23
WIP magic leap support on OSX

shawwn:
This works, but you have to run with

```
DYLD_LIBRARY_PATH=$(pwd)/build/Release node .
```

instead of

```
node .
```

Refs:
https://stackoverflow.com/questions/29773160/node-gyp-link-library-dependencies-at-build-time
https://stackoverflow.com/questions/18748224/how-to-add-dependence-to-static-library-in-binding-gyp-node-gyp-for-node-js-exte/40222639#40222639
https://stackoverflow.com/questions/42512623/how-to-build-nodejs-c-addon-depending-on-a-shared-library-with-relative-locati
https://github.com/bloomberg/blpapi-node/blob/master/binding.gyp


modulesio:
Hm, certainly this can be fixed with an rpath.

Thanks for the progress!

modulesio:
This issue is quite stale with the work in https://github.com/webmixedreality/exokit/pull/362.

PR #74, opened 2018-05-28
Add support for file arguments; add examples folder

shawwn:


index.js
```diff
@@ -1107,6 +1108,9 @@ const _start = () => {
     if (url === '.') {
       console.warn('NOTE: You ran `exokit . <url>`\n(Did you mean to run `node . <url>` or `exokit <url>` instead?)')
     }
+    if (url && url.indexOf('://') < 0) {
```
modulesio: If the user passes in a url in a query string this is broken, isn't it?

shawwn: Yes, good catch.

modulesio:
Thanks!

PR #77, opened 2018-05-29
Add missing gl.getUniform method

shawwn:
This fixes http://stack.gl/gl-shader/

### DIFF
--- a/deps/exokit-bindings/util/include/defines.h
+++ b/deps/exokit-bindings/util/include/defines.h
@@ -24,4 +24,23 @@ private:
 Local<Array> pointerToArray(void *ptr);
 void *arrayToPointer(Local<Array> array);
 
+template <typename T> struct V8TypedArrayTraits;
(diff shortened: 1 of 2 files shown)


modulesio:
Ok so this looks good, but our getUniform implementation does not: it assumes array-y data type for the uniform. Per the mdn spec (https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getUniform), shouldn’t we be returning different type based on the internal type of the uniform by calling eg https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGetActiveUniform.xhtml?

modulesio:
Basically this PR looks good but it seems to reveal a deeper conformance bug. Do we want to merge and open an issue for the conformance or just fix it in one swoop here?

shawwn:
Yeah, good point. I'm not sure whether it's best to merge this or to leave it out until the full solution is implemented.

modulesio:
Thanks for the extra work!

PR #81, opened 2018-05-31
Define nativeBindings.nativeVideo.Video.getDevices()

shawwn:


deps/exokit-bindings/videocontext/include/VideoCommon.h
```diff
@@ -0,0 +1,57 @@
...
+#define EXO_USING_V4L 1
+#endif
+
+#include <string>
+#include <vector>
+#include <sstream>
+
+namespace ffmpeg {
+
+  static int Abs(int a) {
```
modulesio: Could we not remplate this?

deps/exokit-bindings/videocontext/include/VideoMode.h
```diff
@@ -0,0 +1,43 @@
+#ifndef _HTML_VIDEO_MODE_ELEMENT_H_
```
modulesio: What's a "video mode element"?

deps/exokit-bindings/videocontext/src/win/VideoMode_win.cpp
```diff
@@ -0,0 +1,202 @@
+#include "VideoMode.h"
```
modulesio: This whole file should probably be `#if _WIN32`.

binding.gyp
```diff
@@ -195,6 +196,7 @@
             '<!@(ls -1 deps/exokit-bindings/webglcontext/src/*.cc)',
             '<!@(ls -1 deps/exokit-bindings/webaudiocontext/src/*.cpp)',
             '<!@(ls -1 deps/exokit-bindings/videocontext/src/*.cpp)',
+            '<!@(ls -1 deps/exokit-bindings/videocontext/src/linux/*.cpp)',
```
modulesio: I guess these don't exist but will?

deps/exokit-bindings/videocontext/include/VideoCommon.h
modulesio: @shawwn yeah I don't get the point of the type specifications when these could be type templates.

deps/exokit-bindings/videocontext/include/VideoMode.h
modulesio: @shawwn if this is not an element (which is a well-defined term) then the include guard should not refer to it as such.

PR #82, opened 2018-05-31
Initial implementation of getUserMedia for video

shawwn:


core.js
```diff
@@ -2761,7 +2761,190 @@ HTMLAudioElement.HAVE_METADATA = HTMLMediaElement.HAVE_METADATA;
 HTMLAudioElement.HAVE_CURRENT_DATA = HTMLMediaElement.HAVE_CURRENT_DATA;
 HTMLAudioElement.HAVE_FUTURE_DATA = HTMLMediaElement.HAVE_FUTURE_DATA;
 HTMLAudioElement.HAVE_ENOUGH_DATA = HTMLMediaElement.HAVE_ENOUGH_DATA;
+class Video {
+  static getDevices() {
+    return [];
+  }
+}
+
+class MediaDeviceKind {
```
modulesio: Maybe this should just be an object?

core.js
```diff
@@ -2761,7 +2761,190 @@ HTMLAudioElement.HAVE_METADATA = HTMLMediaElement.HAVE_METADATA;
...
+}
+MediaDeviceKind.audioInput = "audioinput";
```
modulesio: Looks like these don't exist yet, so is there a point to this enum for now?

core.js
```diff
@@ -2761,7 +2761,190 @@ HTMLAudioElement.HAVE_METADATA = HTMLMediaElement.HAVE_METADATA;
...
+}
+
+class OverconstrainedError extends Error {
+  constructor(constraint, message) {
+    super(message);
+    this.constraint = constraint;
+  }
+}
+
+const Infinity = 1 / 0;
```
modulesio: I think this is already a JS global constant...

core.js
```diff
@@ -2761,7 +2761,190 @@ HTMLAudioElement.HAVE_METADATA = HTMLMediaElement.HAVE_METADATA;
...
+  return true;
+}
+
+const _isConstraintApplicable = (type, name) => {
+  // TODO
+  return true;
+}
+
+class OverconstrainedError extends Error {
+  constructor(constraint, message) {
```
modulesio: Maybe args the other way around?

core.js
```diff
@@ -2761,7 +2761,190 @@ HTMLAudioElement.HAVE_METADATA = HTMLMediaElement.HAVE_METADATA;
...
+      this.deviceId = device.id;
+      this.groupId = device.id;
+      this.label = device.name;
+    } else {
+      throw new Error("not implemented");
+    }
+  }
+}
+
+class MediaStreamTrack {
```
modulesio: So there is [already a `MediaStreamTrack`](https://github.com/webmixedreality/exokit/blob/master/deps/exokit-bindings/webaudiocontext/src/MediaStreamTrack.cpp#L5) in the C++. How does that reconcile with this design where it's implemented in JS as well?

core.js
```diff
@@ -2782,13 +2972,42 @@ class HTMLVideoElement extends HTMLMediaElement {
...
+    return new DOMRect(0, 0, this.width, this.height);
+  }
+
+  get data() {
+    return this.video ? this.video.data : null;
+  }
+  set data(data) {}
+
+  play() {
+    if (this.video)
```
modulesio: Curlification? That's kind of the standard formatting scheme in the rest of the code base.

core.js
```diff
@@ -2761,7 +2761,190 @@ HTMLAudioElement.HAVE_METADATA = HTMLMediaElement.HAVE_METADATA;
 HTMLAudioElement.HAVE_CURRENT_DATA = HTMLMediaElement.HAVE_CURRENT_DATA;
 HTMLAudioElement.HAVE_FUTURE_DATA = HTMLMediaElement.HAVE_FUTURE_DATA;
 HTMLAudioElement.HAVE_ENOUGH_DATA = HTMLMediaElement.HAVE_ENOUGH_DATA;
+class Video {
```
modulesio: Is this best called "Video" or more like "MediaDevices"?

core.js
shawwn: It's from the spec.

core.js
shawwn: It's from the spec.

core.js
shawwn: It's probably better to implement as much of the spec as possible in JS.

core.js
shawwn: One benefit of calling it Video is that there are many points in the code where you want to iterate over only the video devices, and not e.g. audio inputs. MediaDevices would blur that line.

core.js
modulesio: The spec says it's a class?

core.js
modulesio: My bad, you are correct!

core.js
shawwn: The spec says it's an enum, but that's not a thing in JS. We can redefine it as a plain object if you want.

core.js
modulesio: Either way, it doesn't seem right to have both implementations. What's the design plan there?

core.js
modulesio: Yeah I can see that. But there is `Image` and `Audio`, which are aliases for `HTMLImageElement` and `HTMLAudioElement`. Presumably `Video` would be an alias for `HTMLVideoElement`?

In which case the relationship to media devices probably wouldn't make sense.

core.js
shawwn: I think it makes the most sense to implement MediaStreamTrack in JS, and then expose the low-level APIs via nativeBindings. It's better to do it that way because the selection algorithm is quite complex and better suited to JS. And once you've selected the various characteristics the user wants to use, it's straightforward to pass it along to the native class.

core.js
modulesio: Yeah I think the spec means it as a more abstract concept. I don't think that implies it should be constructible as a class.

I prefer the Object approach 👍.

core.js
shawwn: Honestly the only thing that we need from `Video` is `getMediaDevices()`, which can go pretty much anywhere. The only reason it was named this is because it's called `Video` in the nativeBindings. Feel free to suggest whatever changes you want.

core.js
shawwn: Done

core.js
modulesio: Rename to `MediaStream`?

core.js
modulesio: I can get behind that, but in that case we shouldn't have the MediaStreamTrack class doubly implemented once in JS here and once in the audio subsystem.

modulesio: That is, one of these should probably be deleted and merged into the other.

modulesio:
Closing due to being outdated.

PR #86, opened 2018-06-02
Define VideoDevice

shawwn:


deps/exokit-bindings/videocontext/src/Video.cpp
```diff
@@ -544,6 +550,169 @@ FrameStatus Video::advanceToFrameAt(double timestamp) {
   return status;
 }
 
+
+VideoDevice::VideoDevice() : dataDirty(true), dev(NULL) {
```
modulesio: `nullptr`? Doesn't matter too much.

deps/exokit-bindings/videocontext/src/VideoMode.cpp
```diff
@@ -0,0 +1,204 @@
...
+#endif
+
+#ifdef EXO_OS_OSX
+  if ((iformat = av_find_input_format("avfoundation")))
+      return iformat;
+  if ((iformat = av_find_input_format("qtkit")))
+      return iformat;
+#endif
+
+  fprintf(stderr, "No valid input format found\n");
```
modulesio: Do we want to be spamming this here? Maybe throw instead?

deps/exokit-bindings/videocontext/src/VideoMode.cpp
```diff
@@ -0,0 +1,204 @@
...
+class VideoDeviceImpl : public IVideoDevice
+{
+public:
+  AVFormatContext* pFormatCtx;
+  int videoStream;
+  AVCodec* pCodec;
+  AVFrame* pFrame;
+  AVFrame* pFrameRGB;
+  AVPacket packet;
+  mutable bool dataDirty;
```
modulesio: Why `mutable`?

deps/exokit-bindings/videocontext/src/VideoMode.cpp
```diff
@@ -0,0 +1,204 @@
...
+      return iformat;
+  if ((iformat = av_find_input_format("qtkit")))
+      return iformat;
+#endif
+
+  fprintf(stderr, "No valid input format found\n");
+  return NULL;
+}
+
+class VideoDeviceImpl : public IVideoDevice
```
modulesio: Class def in the `.cpp`? Can we move it to the `.h`, or even its own `.h`?

deps/exokit-bindings/videocontext/include/VideoMode.h
```diff
@@ -4,13 +4,21 @@
 #include "VideoCommon.h"
 
 namespace ffmpeg {
+class IVideoDevice {
```
modulesio: I don't understand the purpose of this interface. It seems there is only one kind of video device, and we could get by with only one `VideoDevce` and no vtables.

deps/exokit-bindings/videocontext/include/Video.h
```diff
@@ -1,6 +1,7 @@
 #ifndef _HTML_VIDEO_ELEMENT_H_
 #define _HTML_VIDEO_ELEMENT_H_
 
+
```
modulesio: Extra space.

deps/exokit-bindings/videocontext/include/VideoMode.h
```diff
@@ -3,14 +3,186 @@
...
+public:
+  AVFormatContext* pFormatCtx;
+  int videoStream;
+  AVCodec* pCodec;
+  AVFrame* pFrame;
+  AVFrame* pFrameRGB;
+  AVPacket packet;
+
+
+  VideoCamera(AVFormatContext* formatContext, AVCodec* codec, int videoStream)
```
modulesio: Couldn’t these definitions be in the cpp? That’s the convention everywhere else I think.

PR #89, opened 2018-06-04
Fix OSX build

shawwn:
I don't know why this doesn't error on Travis, but on my High Sierra
MBP there was a build error due to AVMediaType being defined in
avutil.h and in AVFoundation.


deps/exokit-bindings/videocontext/src/Video.cpp
```diff
@@ -1,5 +1,6 @@
 #include <Video.h>
 #include "VideoMode.h"
+#include "VideoCamera.h"
```
modulesio: Can we clean these up re angle and non-angle? I guess it was already inconsistent but this PR adds additional inconsistency.

PR #94, opened 2018-06-04
Improve configurator

shawwn:
- Activate VR using vrdisplayactivate
- Use vive controller models
- Fix FOV calculation
- Store points in local storage
- Render controller models where the user's eyes are


examples/003_configure.html
```diff
@@ -237,49 +287,67 @@
...
-                 );
-                 camera.updateMatrixWorld();
-                 matrixWorld.copy(camera.matrixWorldInverse);
+               if (!points.init) {
+                 points.push(controllerMesh.position.clone());
+                 console.log(`point ${points.length}: ${JSON.stringify(points[points.length - 1])}`);
+               }
+
+               const _configureCamera = (camera, points) => {
+                 P = points[0]; // real-world camera position
```
modulesio: Is this using globals?

examples/003_configure.html
```diff
@@ -74,8 +75,6 @@
   </script>
   <script>
 
-			let camera, matrixWorld, projectionMatrix, scene, renderer, video, controllerMeshes, lastPresseds, points, presenting;
```
modulesio: What’s the point of globalizing here?

examples/003_configure.html
```diff
@@ -196,11 +230,27 @@
...
-                        projectionMatrix.toArray(updateSpec.frameData.leftProjectionMatrix);
-
-                        matrixWorld.toArray(updateSpec.frameData.rightViewMatrix);
-                        projectionMatrix.toArray(updateSpec.frameData.rightProjectionMatrix);
+                        const {frameData} = updateSpec;
+
+                        localMatrix.fromArray(frameData.leftViewMatrix);
+                        localMatrix.getInverse(localMatrix);
+                        localMatrix.decompose(localVector, localQuaternion, localVector2);
+                        controllerMeshes[2].quaternion.fromArray(frameData.pose.orientation);
```
modulesio: We don’t need to set this every frame right? Only when actually set.

examples/003_configure.html
```diff
@@ -3,6 +3,7 @@
 
 		<video id="video" autoplay style="display:none"></video>
   <script src="https://cdnjs.cloudflare.com/ajax/libs/three.js/93/three.js"></script>
+  <script src="js/OBJLoader.js"></script>
```
modulesio: We pull three from cdn so why not this?

modulesio:
This was cleaned up and merged on or about https://github.com/webmixedreality/exokit/commit/7aa5a062049f64500d90d2d98cdc9b1ac179327a.

Thanks!

PR #96, opened 2018-06-05
Render window separately from HMD

shawwn:


core.js
```diff
@@ -3267,11 +3268,14 @@ const _runJavascript = (jsString, window, filename = 'script', lineOffset = 0, c
 };
 
 let rafCbs = [];
-const tickAnimationFrame = () => {
+const tickAnimationFrame = which => {
   if (rafCbs.length > 0) {
     const localRafCbs = rafCbs.slice();
 
     const _handleRaf = localRafCb => {
+      if (which && !which(localRafCb[whichSymbol])) {
```
modulesio: `which` should probably never be false, so there's no point to check.

modulesio: Also, I would make `which` a constant to compare against, to optimize. This is one of the hottest paths in the codebase.

core.js
```diff
@@ -3779,11 +3783,15 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
       }
     }
   };
-  window.requestAnimationFrame = fn => {
-    fn[windowSymbol] = window;
-    rafCbs.push(fn);
-    return fn;
-  };
+  const _requestAnimationFrame = which => {
```
modulesio: Could be a single double-arrow to clean up the closure.

index.js
```diff
@@ -522,6 +522,46 @@ const _bindWindow = (window, newWindowCb) => {
     console.warn('got error', err);
   });
 
+  const _blit = ({submit, swap, update}) => {
```
modulesio: I'd make this take three args instead of an object, as an optimization.

index.js
```diff
@@ -931,57 +971,42 @@ const _bindWindow = (window, newWindowCb) => {
         }
       }
     }
-    window.tickAnimationFrame();
```
modulesio: This doesn't seem like it's related to this PR, but it's a good change.

PR #99, opened 2018-06-05
Implement new Audio('some-source-url')

shawwn:


core.js
```diff
@@ -3334,9 +3334,12 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
   })(HTMLImageElement);
   const HTMLAudioElementBound = (Old => class HTMLAudioElement extends Old {
     constructor() {
-      super(...arguments);
-
+      let [attrs, ...args] = [...arguments]
```
modulesio: Instead of this fiddling can we do
```
constructor(src) {
if (typeof src === 'string') {
  const audio = new HTMLAudioElementBound();
  audio.setAttribute('src', src);
  return audio;
} else {
...

It seems cleaner and more futureproof.

shawwn: Good call. Updated.

PR #101, opened 2018-06-07
Fix build error

shawwn:
For some reason, our appveyor build process was failing with an error about `std::back_inserter` being undefined. This appears to fix it.

Not sure why you're not running into the same problem.

### DIFF
--- a/deps/exokit-bindings/videocontext/src/VideoMode.cpp
+++ b/deps/exokit-bindings/videocontext/src/VideoMode.cpp
@@ -1,6 +1,9 @@
 #include <VideoMode.h>
 #include <VideoCamera.h>
 
+#include <algorithm>
+#include <iterator>
+
 namespace ffmpeg {
 
 template<typename Out>


modulesio:
This doesn't appear to be a problem with the current exokit CI, so it would be helpful to get a more thorough explanation of the problem and how this fixes it rather than adding mystery code 🔍.

shawwn:
I'm pretty sure this is because of no MAGICLEAP environment var. Niltree doesn't need it anyway.

shawwn:
When MAGICLEAP is set, these headers are included in bindings.h:

```
8:#ifndef _MAGICLEAP_H_
9:#define _MAGICLEAP_H_
10-
11-#include <v8.h>
12-#include <node.h>
13-#include <nan/nan.h>
14-#include <defines.h>
15-#include <thread>
16-#include <mutex>
17-#include <condition_variable>
18-#include <glfw.h>
19-#include <ml_graphics.h>
20-#include <ml_head_tracking.h>
21-#include <ml_perception.h>
22-#include <ml_planes.h>
23-#include <ml_meshing.h>
24-#include <ml_input.h>
25-#include <ml_gesture.h>
26-#include <ml_occlusion.h>
27-#include <ml_lifecycle.h>
28-#include <ml_logging.h>
```

shawwn:
I think if you build without MAGICLEAP set, you'll see the same error.

modulesio:
I have no such error locally with https://github.com/webmixedreality/exokit/commit/9728af02ab4f5347dd18bf39192bab35dc278c32.

Tried both with and without `MAGICLEAP` set, so this sounds tooling-specific.

What Visual Studio is being used by `node-gyp` for the repro?

shawwn:
Hm. I didn't repro it locally, only on appveyor. I assumed MAGICLEAP was the cause.

If I find time I'll investigate this more thoroughly.

modulesio:
This was merged in https://github.com/webmixedreality/exokit/commit/38917b637c719ec1dc013b68bb7a1e92a811ae37.

PR #105, opened 2018-06-08
Fix startup script for workers

shawwn:


core.js
```diff
@@ -3399,7 +3399,7 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
     const WebSocket = require('ws/lib/websocket');
     const {XMLHttpRequest} = require('window-xhr');
     const XHRUtils = require('window-xhr/lib/utils');
-    ${!args.require ? 'window.require = undefined;' : ''}
```
modulesio: Maybe this should just be `global`? Does that work?

PR #111, opened 2018-06-10
Separate device/window render

shawwn:
This PR lets sites render something to the window and something
different to the HMD.

You can test using the following sites. (Press P to load into VR.)

A standard VR site: http://emkolar.ninja

A site that renders differently in the HMD than the window: https://files.webmr.io/_files/shawnpresser@gmail.com/sloth-necessary-29/leetsaber/index.html

NOTE: if a user site uses window.RAF instead of device.RAF while VR is enabled, the HMD will be one frame behind, which is pretty jarring. You can see this for yourself in the first site.

I'm not sure how to do better than this. Thoughts?

### DIFF
--- a/core.js
+++ b/core.js
@@ -3409,11 +3409,11 @@ const tickAnimationFrame = () => {
   }
 };
 
-const _makeRequestAnimationFrame = window => (fn, priority = 0) => {
+const _makeRequestAnimationFrame = (window, prio = 0) => (fn, priority = prio) => {
   fn[windowSymbol] = window;
   fn[prioritySymbol] = priority;
   rafCbs.push(fn);
-  rafCbs.sort((a, b) => b[prioritySymbol] - a[prioritySymbol]);
+  rafCbs.sort((a, b) => a[prioritySymbol] - b[prioritySymbol]);
   return fn;
 };
 const _getFakeVrDisplay = window => window[mrDisplaysSymbol].fakeVrDisplay;
@@ -3913,7 +3913,7 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
       }
     }
   };
-  window.requestAnimationFrame = _makeRequestAnimationFrame(window);
+  window.requestAnimationFrame = _makeRequestAnimationFrame(window, 2);
   window.cancelAnimationFrame = fn => {
     const index = rafCbs.indexOf(fn);
     if (index !== -1) {
--- a/index.js
+++ b/index.js
@@ -527,35 +527,37 @@ const _bindWindow = (window, newWindowCb) => {
     console.warn('got error', err);
   });
 
-  const _blit = () => {
+  const _blit = toDevice => {
     for (let i = 0; i < contexts.length; i++) {
       const context = contexts[i];
 
       if (context.isDirty()) {
         const windowHandle = context.getWindowHandle();
+        nativeWindow.setCurrentWindowContext(windowHandle);
 
-        if (nativeWindow.isVisible(windowHandle) || vrPresentState.glContext === context || mlGlContext === context) {
-          if (vrPresentState.glContext === context) {
-            nativeWindow.setCurrentWindowContext(windowHandle);
-
-            nativeWindow.blitFrameBuffer(context, vrPresentState.msFbo, vrPresentState.fbo, renderWidth * 2, renderHeight, renderWidth * 2, renderHeight, true, false, false);
-            vrPresentState.compositor.Submit(context, vrPresentState.tex);
-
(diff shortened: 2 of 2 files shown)


core.js
```diff
@@ -3913,7 +3913,7 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
       }
     }
   };
-  window.requestAnimationFrame = _makeRequestAnimationFrame(window);
+  window.requestAnimationFrame = _makeRequestAnimationFrame(window, 2);
```
modulesio: Can this be rebased to zero? So the user raf is zero and anything before that is negative?

shawwn: Not easily. The user RAF has to come after the device RAF, and rebasing the user RAF to zero means either defaulting the parameter to a negative value or passing in a specific constant in N different places. (Where N is about 4.)

index.js
```diff
@@ -975,25 +977,48 @@ const _bindWindow = (window, newWindowCb) => {
...
+        _blit(true);
+        if (args.performance) {
+          const now = Date.now();
+          const diff = now - timestamps.last;
+          timestamps.submit += diff;
+          timestamps.total += diff;
+          timestamps.last = now;
+        }
+      },
+      1);
```
modulesio: So overall this seems like the frame flow is:

1. Update matrices
2. Blit do HMD
3. Run user rafs
4. Bit to window

This doesn't seem right to me. Should the user code always run first? Otherwise as mentioned the HMD blit will be one frame behind.

shawwn: The user code can't run first unless you figure out some way of deciding whether the HMD needs to blit its split screen to the window.

Right now the HMD blits to the window, then the window can overwrite that blit with a different render. If you reverse it, the HMD blit will stomp on the window render.

I'm not sure there's a good way in general to detect whether a site is trying to do separate renders for window vs device.

Ideas are welcome, but it seems like a tough problem.

core.js
modulesio: Hm, I really don't like using these constants in general. It makes it essentially impossible to use the raf sorting feature if the core code is defining its own constants.

I'd prefer to either get rid of rafs inside of this code, or define basis constant for the frame times so user code can at least do a `+`/`-` off of them.

shawwn: Sure, think over a design that makes sense and let me know what you decide.

index.js
modulesio: > I'm not sure there's a good way in general to detect whether a site is trying to do separate renders for window vs device.

Well at this point in the code we already know that don't we? We know whether the user has requested device or window rafs based on which they called. This was previously (and can still) be flagged at raf registration time.

> Right now the HMD blits to the window, then the window can overwrite that blit with a different render. If you reverse it, the HMD blit will stomp on the window render.

So it seems the order should be:

1. Update matrices
1. Run device rafs
1. If there were device rafs, blit to device AND back to window, otherwise no blits
1. Run window rafs
1. If there were window rafs, blit to window

shawwn: And how do we determine whether there were device rafs? :) That was the functionality you removed.

modulesio: Indeed, but that can easily be put back.

core.js
```diff
@@ -3409,11 +3409,11 @@ const tickAnimationFrame = () => {
   }
 };
 
-const _makeRequestAnimationFrame = window => (fn, priority = 0) => {
+const _makeRequestAnimationFrame = (window, prio = 0) => (fn, priority = prio) => {
```
modulesio: This thing can take a device string and decorate it on the `fn` via `Symbol`.

index.js
modulesio: Actually I think this is a bit more complicated since you probably don't want to destructively blit in `3` if there was no context update. That's what `gl.isDirty()` is for (it's tracked automatically at the C++ layer).

core.js
shawwn: In addition to the prio? That might work.

index.js
shawwn: Yes, that's what I meant by "Certain prios should only run if certain other prios run." Not so crazy after all!

Unfortunately splitting it up by prio makes that logic difficult. Ideas?

modulesio: I don't think any of the above code should use raf itself. Its job is to execute user rafs, but it shouldn't need to use them. That gets rid of the problem of dictating over raf prios.

Just straight ifs and function calls should do it I think.

shawwn: If the blits don't become RAFs, what was the point of the prio refactor with sorting? I thought blits were supposed to be RAFs?

modulesio: The prio refactor was to support the needs of `exokit-home`, which is for all purposes effectively just a user site. So it's allowed to dabble in and leverage the raf sorting feature.

I don't think the same goes for the core -- it's not a user site and it has enough context to not need to resort on raf sorting.

modulesio:
I think this PR is stale and would need a rethink.

PR #122, opened 2018-06-13
Add support for XBox 360 and PS4 controllers

shawwn:


core.js
```diff
@@ -3854,6 +3854,7 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
     }
   };
   // window.Buffer = Buffer; // XXX non-standard
+  window.G = global;
```
modulesio: What does this have to do with gamepads?

deps/openvr/src/ivrsystem.cpp
```diff
@@ -841,12 +906,42 @@ NAN_METHOD(IVRSystem::TriggerHapticPulse)
...
+  uint32_t unAxisId = info[1]->Int32Value();
+  double dValue  = info[2]->NumberValue();
+  unsigned short usDurationMicroSec = info[3]->Uint32Value();
+  if (unAxisId >= 2) {
+    Nan::ThrowTypeError("Argument[1] must be 0 or 1.");
+    return;
+  }
 
-  obj->self_->TriggerHapticPulse(unControllerDeviceIndex, unAxisId, usDurationMicroSec);
+  if (iIndex < 0) {
```
modulesio: This seems very much a hack. Since this code is effectively bug openvr, can we move it to an entirely new dir?

Deciding this branch is probably best done higher in js, not in “openvr” (which it is apparently not).

index.js
```diff
@@ -200,6 +200,16 @@ const _normalizeMatrixArray = float32Array => {
...
 };
 
+const gamepadState = {
+  system: null,
+  isPresenting: true,
+}
+
+{
+  const vrContext = nativeVr.getContext();
+  gamepadState.system = nativeVr.VR_Init(nativeVr.EVRApplicationType.Scene);
```
modulesio: Won’t this auto boot vr for everything? If that’s true then I think this isn’t a good change in behavior.

index.js
```diff
@@ -541,7 +551,7 @@ const _bindWindow = (window, newWindowCb) => {
       if (context.isDirty()) {
         const windowHandle = context.getWindowHandle();
         nativeWindow.setCurrentWindowContext(windowHandle);
-        context.flush();
+        context.flush(); // TODO: Get rid of this; it causes a pipeline stall.
```
modulesio: Flush stalls? I thought finish does, not flush. But it’s a hint to OS X to commit raleslith tab frame buffers in sync.

It also seems unrelated to this pr.

core.js
shawwn: Whoops. I use that for debugging. I’ll nuke it.

It’s so handy that you might want to flag it.

deps/openvr/src/ivrsystem.cpp
shawwn: High praise. 

I disagree. I think that openvr is designed to support exactly this case, and their API is thoughtful and general. Exposing extra controllers for free is something everyone would want — why force devs to do extra work to support controllers, joysticks, steering wheels, and foot pedals? The same API can expose them all automatically.

index.js
shawwn: I’ll check. But if so, we can just return a stub OpenVR instance without actually calling init.

index.js
shawwn: It seemed like I was getting higher FPS without it. Will do more testing.

deps/openvr/src/ivrsystem.cpp
modulesio: I agree the API is good and can be used. I just disagree that directx is a part of binding to openvr dlls, so the files could use a restructuring.

index.js
modulesio: I think it will boot immediately.

> But if so, we can just return a stub OpenVR instance without actually calling init.

Yes I think that's fine.

modulesio:
Closing due to code staleness -- however, this is probably something we could pick up again if there was a feature champion for it.

PR #127, opened 2018-06-15
testRun.sh, code review for CI testing.

ChrisEddy:
Hey guys, this is my bash script i've figured out.

So far it can:

1.) Open exokit.sh - h 
2.) Log console stdout and stderr to a testRunLog.txt 
3.) Provides exit codes 

PID is broken and does not kill process yet. Been trying to use pgrep.

This is for review and suggestions.


testRun.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+SCRIPT_PATH="./exokit.sh"
+
+# Execute ./exokit.sh -h and log the console output to testRunLog.txt,
+# stdout and stderr included
+"$SCRIPT_PATH" -h >testRunLog.txt 2>&1 &
+echo "Exokit loaded, Exit Code: " $?
+
+sleep 5s
```
modulesio: Instead of this, we could just make exokit have a flag that makes it render once and then exit success or failure depending on whether it thinks it drew a frame.

testRun.sh
```diff
@@ -0,0 +1,20 @@
...
+# Get process ID of exokit and kill it. (does not work )
+PID=$(pgrep -u root exokit.cmd)
+$(pkill -SIGKILL $PID)
+echo "Process killed, exit Code: " $?
+
+echo "Finished test run of ./exokit.sh -h"
+
+
+# keep console open for user to read or else goes away to quick
+read
```
modulesio: I probs wouldn't keep this open. Since this is for CI/testing I think it's ok to require this test to be run in a parent console.

Also if we hang like this then it will just hang CI and always fail.

ChrisEddy: yes for sure I will take this out... more for dev use on my own machine.

PR #132, opened 2018-06-20
Fix iframe rendering

shawwn:


modulesio:
After more research, this PR does seem correct. 👍 

The top window binding effectively handles the bindings for all windows, including children. Therefore filtering out the blit to only the top window is the wrong thing to do.

Notably, the destroy event passes the target window as part of the event, so it's not a problem that its destruction is handled by the top level window only, because the window to destroy is sourced from the event property.

Thanks!

PR #133, opened 2018-06-20
Make vm-one optional

shawwn:


modulesio:
Hm, I’m not sure if it’s a matter of options. The only reason vm-one exists is that V8 couldn’t optimize typed arrays/arraybuffers due to species mismatch, which made the core intended use case (90 FPS MR) unworkable for common site data flows.

Therefore it seems we either need to:
- redefine the goals of what we’re supporting
- check that vm-one is no longer needed and we can delete it, or
- vm-one is still needed for perf and this PR effectively adds a known problematic mode, which I would be somewhat reluctant to support

Thoughts?

shawwn:
> check that vm-one is no longer needed and we can delete it

How would I do this?

shawwn:
Found a good benchmark.

With vm-one:
```
>node . -p "https://threejs.org/examples/webgl_shadowmap_performance.html"
Lifecycle library is not connected.
THREE.WebGLRenderer 93
THREE.JSONLoader: "morphColors" no longer supported. Using them as face colors.
50 FPS | 1177ms idle | 1ms wait | 0ms prepare | 3ms events | 0ms media | 611ms user | 21ms submit
49 FPS | 138ms idle | 1ms wait | 0ms prepare | 4ms events | 1ms media | 1668ms user | 29ms submit
51 FPS | 148ms idle | 0ms wait | 0ms prepare | 0ms events | 0ms media | 1595ms user | 27ms submit
57 FPS | 132ms idle | 0ms wait | 0ms prepare | 1ms events | 0ms media | 1417ms user | 25ms submit
52 FPS | 140ms idle | 0ms wait | 0ms prepare | 5ms events | 0ms media | 1565ms user | 28ms submit
48 FPS | 143ms idle | 0ms wait | 0ms prepare | 4ms events | 0ms media | 1702ms user | 30ms submit
55 FPS | 143ms idle | 0ms wait | 0ms prepare | 2ms events | 0ms media | 1451ms user | 32ms submit
57 FPS | 149ms idle | 0ms wait | 0ms prepare | 2ms events | 1ms media | 1407ms user | 25ms submit
Lifecycle library is not connected.
49 FPS | 144ms idle | 0ms wait | 0ms prepare | 3ms events | 0ms media | 1649ms user | 32ms submit
53 FPS | 139ms idle | 0ms wait | 0ms prepare | 1ms events | 0ms media | 1539ms user | 23ms submit
```

Without vm-one:

```
>node . -p "https://threejs.org/examples/webgl_shadowmap_performance.html"
Lifecycle library is not connected.
THREE.WebGLRenderer 93
THREE.JSONLoader: "morphColors" no longer supported. Using them as face colors.
47 FPS | 1101ms idle | 1ms wait | 0ms prepare | 2ms events | 0ms media | 791ms user | 17ms submit
32 FPS | 144ms idle | 0ms wait | 0ms prepare | 3ms events | 0ms media | 2604ms user | 29ms submit
35 FPS | 139ms idle | 1ms wait | 0ms prepare | 0ms events | 1ms media | 2397ms user | 31ms submit
35 FPS | 152ms idle | 0ms wait | 0ms prepare | 2ms events | 1ms media | 2373ms user | 22ms submit
33 FPS | 143ms idle | 0ms wait | 0ms prepare | 4ms events | 0ms media | 2535ms user | 30ms submit
Lifecycle library is not connected.
34 FPS | 136ms idle | 0ms wait | 0ms prepare | 1ms events | 0ms media | 2469ms user | 24ms submit
33 FPS | 141ms idle | 0ms wait | 0ms prepare | 1ms events | 1ms media | 2550ms user | 24ms submit
38 FPS | 150ms idle | 1ms wait | 0ms prepare | 2ms events | 0ms media | 2219ms user | 25ms submit
```

So it looks like vm-one still gives a sizable performance boost.

Could you explain more about why vm-one is so much faster?

shawwn:
Just so I understand, which part of https://github.com/nodejs/node/blob/0eb8a8cb12ec53dd0ef04bb80d839212a508250c/src/node_contextify.cc#L855 do you think is responsible for the slowdown compared to https://github.com/modulesio/vm-one/blob/a65969f420341fc67de947fef31975c0d1e17d01/main.cpp#L76 ?

modulesio:
> Just so I understand, which part of https://github.com/nodejs/node/blob/0eb8a8cb12ec53dd0ef04bb80d839212a508250c/src/node_contextify.cc#L855 do you think is responsible for the slowdown compared to https://github.com/modulesio/vm-one/blob/a65969f420341fc67de947fef31975c0d1e17d01/main.cpp#L76 ?

It's this one: https://github.com/nodejs/node/blob/0eb8a8cb12ec53dd0ef04bb80d839212a508250c/src/node_contextify.cc#L143

shawwn:
This is necessary for #135.

shawwn:
Actually, it's not enough. The problem is that `yarn add exokit` fails inside of an electron project due to the fact that it tries to build `vm-one`. It would have to be optional at build time, which could only be done by publishing a separate package other than `exokit` (I think..?)

shawwn:
Yeah, I dunno a good solution to this. I started adding some detection code to `preinstall.js` to try to check whether exokit is being installed into electron, and if so, uninstall `vm-one`:

```
// In electron, exokit is installed to app/node_modules/exokit
if (__dirname.endsWith(path.join('app', 'node_modules', 'exokit'))) {
  console.log('Electron detected');
  childProcess.execSync('npm uninstall vm-one --save --ignore-scripts --no-lockfile', {
    cwd: __dirname
  });
}

```

Then I changed `native-bindings.js` like:

```
const useVmOne = require(path.join(__dirname, 'package.json')).dependencies.hasOwnProperty('vm-one');
const vmOne = !useVmOne ? {
  make() {
...
  }
} : require('vm-one');
```

That .. works ... sort of... but it's rather quease-inducing.

modulesio:
Model-gyp can pick up env cars and other build variables. That’s how we detect android/magic leap builds among others.

shawwn:
Yeah, but the issue is that vm-one needs to be an optional --- Ooooh I see, you're saying to put it into vm-one's binding.gyp. Hmmm, I'll try.

PR #140, opened 2018-06-25
Toggle bugsnag with NO_BUGSNAG=1

shawwn:


modulesio:
I'm not sure if this should be encouraged... if there is a problem with bugsnag itself that should probably be fixed.

shawwn:
Being able to turn off telemetry is usually a good idea. Especially when that telemetry is broken.

modulesio:
To be fair, would we argue to fix tests by deleting them? The time we would want telemetry the most seems when there is a break.

shawwn:
It was quite amazing to see the call trace, all the way down to OpenVR.

PR #141, opened 2018-06-25
Initial support for requiring tarballs

shawwn:


core.js
```diff
@@ -16,6 +16,24 @@ const parse5 = require('parse5');
...
+    return e;
+  }
+}
+
+const mkdir = path => {
+  return require('mkdirp').sync(path);
+}
+
+// Nuke site_modules on startup.
+rmrf(SITE_MODULES)
```
modulesio: This should probably not be done synchronously.

core.js
```diff
@@ -16,6 +16,24 @@ const parse5 = require('parse5');
 const fetch = require('window-fetch');
 const {Request, Response, Headers, Blob} = fetch;
 
+const SITE_MODULES = path.join(__dirname, 'site_modules');
+
+const rmrf = path => {
```
modulesio: There are pretty standard modules for doing these.

core.js
```diff
@@ -2894,7 +2916,48 @@ class HTMLScriptElement extends HTMLLoadableElement {
...
+                    const dst = path.join(SITE_MODULES, filename);
+                    fs.writeFileSync(dst, data);
+                    const decompress = require('decompress');
+                    var writer = fs.createWriteStream(dst);
+                    writer.write(Buffer.from(ab));
+                    writer.on('finish', () => {
+                      decompress(dst, SITE_MODULES).then(files => {
+                        const dir = path.join(SITE_MODULES, name);
+                        try {
+                          rmrf(dir);
```
modulesio: This is quite bad for performance. We shouldn't do anything that blocks the main thread, especially on loading. This will basically hang VR depending on the disk speed, beyond what is actually needed to load the file.

core.js
```diff
@@ -2878,6 +2896,10 @@ class HTMLStyleElement extends HTMLLoadableElement {
     return running;
   }
 }
+const _shouldRequire = url => (url.endsWith('.tgz') ||
```
modulesio: This should probably just check the `Content-Type`.

core.js
shawwn: It should, because the alternative is to nest the entirety of core.js inside of a callback. :P

core.js
shawwn: It uses rimraf, which afaik a lot of projects use.

core.js
shawwn: That might be dangerous. A cursory google suggests that .tar.gz files are often misconfigured server-side: https://github.com/haskell/cabal/issues/615

The thing I'm worried about is, .tar.gz isn't actually a "tar-gz", it's "gzipped data". Which happens to be exactly the same type of data that servers usually send down, even if it's just text and not a tarball.

I'm sure there's probably a way to decompress the stream in memory and check whether it's a tarball, but... there's no situation where <script src="foo.tar.gz"></script> means anything other than require it as a tarball.

core.js
shawwn: I think there's some confusion here. `decompess` doesn't block the main thread -- the only thing that does is the `rmrf`, and 99% of the time, `dir` does not exist. The rmrf is there to ensure that it's possible to retry a failed require, which is a very rare case.

core.js
modulesio: But we are already bootin the site async, since it needs to go throught a request. There is no technical reason to not parallelize that.

core.js
modulesio: Can we skip this entirely and if the type is application/require we assume it’s a tarball?

shawwn: Yeah. I wanted to, but for some reason setting `type="application/require"` causes the data not to fetch. Not sure why yet. That's when I switched to filename extension detection.

core.js
shawwn: A more worrying reason is that this will rmrf the `site_modules` whenever a web worker loads. Suffice to say, yes, this needs more thought.

core.js
modulesio: Htmlscriptekement follows spec here. This new type would need to be whitelisted for it to load. That seems correct behavior.

PR #142, opened 2018-06-29
Start xvfb for webgl testing on travis

shawwn:
I don't know whether this will work. This PR is to check whether Travis is happy or not.


modulesio:
Nice, does this do gl calls now? Has that been checked?

modulesio:
This seemed to be on the right track. For posterity is there a technical reason for closing it?

chrisplatorres:
This might be useful to add a "supported XR sites" test in CI that helps maintain a list of sites that are confirmed to open in XR.

modulesio:
Closing due to being outdated.

PR #150, opened 2018-07-03
Add emukit.cmd + installer shortcut

shawwn:


modulesio:
I'm not sure if this is the right place to advertise Emukit; this feels like it should be an indexed link, not a shortcut in the installer...

shawwn:
Fair enough, though I have to tell users to open command prompt and run `C:\exokit\exokit.cmd http://emukit.webmr.io` manually.

PR #151, opened 2018-07-03
start breaking core into modules (fixes #145)

### LINKED ISSUE #145: Break up core.js
Most of the "browser" part of the Exokit browser is in [core.js](https://github.com/webmixedreality/exokit/blob/master/core.js).

There is no good reason for it being contained in one file; it's convenient but it makes the code seem scarier than it is to newcomers.

`core.js` should already have a reasonably coherent layout and factoring; we just need to classify the pieces and break them up into their own files that are `require`d in.

A lot of this is just copy-paste.

### File/directory sections

This is a rough idea for the files/directories we'd probably want:

- Network
- GL
- Audio
- Video
- Canvas
- Scripting contexts
- VR
- XR
- Magic Leap
- Leap Motion
- Events
- DOM
- Observers
- Window
- Document
- Bootstrap
- Runtime loop/queues

ngokevin:
Start breaking core.js:

- DOM
- Document (perhaps goes in DOM)
- Event
- History
- Location
- MutationObserver
- Network
- XR
- constants
- urls
- utils
- symbols

Notes:

- Moved various shared functions and constants to their own modules (e.g., constants, symbols, urls).
- Some shared functions (e.g., `_makeWindow`) not yet moved over to separate modules. I made various stuff `global` in `core.js` for now, and can finish moving over other stuff in a second part.

Tested:

- [x] Home space loads
- [x] A-Painter Loads
- [x] `.go http://tough-meeting.glitch.me/` A-Frame Hello World loads 
- [x] Tested with Vive on Windows, entering VR, navigating to A-Painter, and drawing

Found issues (unrelated to PR):

- Warning to `setMaxEventListeners`.
- When launching Supercraft from CLI and entering VR, error that says "app is already launching". **edit:** might be a weird SteamVR issue.
- Spot the Bot crashes (probably already known).

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -9,6 +9,7 @@ const util = require('util');
 
 const {defaultCanvasSize} = require('./constants');
 const {Event, EventTarget, MouseEvent} = require('./Event');
+const GlobalContext = require('./GlobalContext');
 const symbols = require('./symbols');
 const utils = require('./utils');
 const {_elementGetter, _elementSetter} = require('./utils');
@@ -1273,7 +1274,7 @@ class HTMLStyleElement extends HTMLLoadableElement {
           .then(s => css.parse(s).stylesheet)
           .then(stylesheet => {
             this.stylesheet = stylesheet;
-            styleEpoch++;
+            GlobalContext.styleEpoch++;
             this.dispatchEvent(new Event('load', {target: this}));
           })
           .catch(err => {
@@ -1286,7 +1287,7 @@ class HTMLStyleElement extends HTMLLoadableElement {
         .then(() => css.parse(innerHTML).stylesheet)
         .then(stylesheet => {
           this.stylesheet = stylesheet;
-          styleEpoch++;
+          GlobalContext.styleEpoch++;
           this.dispatchEvent(new Event('load', {target: this}));
         })
         .catch(err => {
@@ -1705,7 +1706,7 @@ class HTMLIFrameElement extends HTMLSrcableElement {
                 baseUrl: url,
                 dataPath: options.dataPath,
               }, parentWindow, parentWindow.top);
-              const contentDocument = _parseDocument(htmlString, contentWindow[symbols.optionsSymbol], contentWindow);
+              const contentDocument = GlobalContext._parseDocument(htmlString, contentWindow[symbols.optionsSymbol], contentWindow);
               contentDocument.hidden = this.hidden;
 
               contentWindow.document = contentDocument;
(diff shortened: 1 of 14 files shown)


shawwn:
Note that we'll lose all `git blame` and `git log` history for core.js, which is quite useful for seeing how exokit's design has evolved.

(There are ways around this, but that's the default behavior.)

modulesio:
I think clearly seeing the design as it is now is more important than how it got there. Especially since this breakup was mostly just one author on the `blame` anyway.

ngokevin:
Yeah, `git blame` would need a pitstop by this commit if they want to go past this point in time. And we'll get to do like `git log XR.js` to see the history of a module / area of code too. That should be pretty useful.

src/Document.js
```diff
@@ -0,0 +1,67 @@
...
+    if (topDocument[symbols.fullscreenElementSymbol] !== null) {
+      topDocument[symbols.fullscreenElementSymbol] = null;
+
+      process.nextTick(() => {
+        topDocument._emit('fullscreenchange');
+      });
+    }
+  }
+}
+global.Document = Document;
```
shawwn: Breaking apart core.js makes it easier for this type of bug to slip through.

ngokevin: Yeah, it's a temporary step. I'm using `global` until we can break the rest of `core.js` into modules because there would be circular dependencies in the meantime, and I wanted to do it piecewise.

src/Document.js
```diff
@@ -0,0 +1,69 @@
...
+      process.nextTick(() => {
+        topDocument._emit('fullscreenchange');
+      });
+    }
+  }
+}
+module.exports.Document = Document;
+
+// FIXME: Temporary until we can refactor out into modules more and not have circular dependencies.
+global._Document = Document;
```
modulesio: This isn't terrible, but we could avoid leaking globals with nearly the same code:

```
const _globalContext = require('./global-context.js'); // empty file
/// ...
_globalContext.Document = Document;
```

shawwn:
I just tested this against PhysiJS:

image

Everything seems to work fine, so this PR looks good to me. 👍

src/Document.js
ngokevin: Alright, I'll set up a temporary global module.

ngokevin: Okay, created a GlobalContext module.

src/DOM.js
```diff
@@ -0,0 +1,1998 @@
...
+module.exports.HTMLMediaElement = HTMLMediaElement;
+
+class HTMLSourceElement extends HTMLSrcableElement {
+  constructor(attrs = [], value = '', location = null) {
+    super('SOURCE', attrs, value, location);
+  }
+}
+module.exports.HTMLSourceElement = HTMLSourceElement;
+
+class HTMLImageElement extends HTMLSrcableElement {
```
modulesio: Posterity: these stubs are probably pointless.

The original idea was to have available a pure-JS impl of DOM with no native bindings, but... afaik nobody's ever used that feature, and these stubs aren't maintained.

modulesio:
Thanks, LGTM!

ngokevin:
Cool, thanks @shawwn + @modulesio!

PR #152, opened 2018-07-03
Linux WebXR: do not include libexpat

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/147.

The bug addressed here is that the Linux Skia build included a copy of libexpat, which was not the same version as the libexpat in various graphics drivers, causing xml parsing crashes without an `LD_PRELOAD`.

I think this is the last of the changes needed for Exokit to work on Arch Linux, provided the deps are installed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "minimist": "^1.2.0",
     "mkdirp": "^0.5.1",
     "native-audio-deps": "0.0.32",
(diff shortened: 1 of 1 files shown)


SketchStick:
With this change I was able to run the Three.js WebVR demos without any LD_PRELOAD's.

PR #153, opened 2018-07-03
start unit test with mocha/sinon/chai

ngokevin:
Just set up a basic unit test to start.

Initial unit test imports core with native bindings, makes a window, and parses a document.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- '9'
+- '10'
 env:
   matrix:
     secure: CTziiKsHYCjdRYftawXT7oBMFm3nOQNRc+ZFv0rUxTaGkw9ImwuxBv6yqMGh6YfykIe94aFNwV+oN8vbuG9NZ56xotAW7mt4r/eCQRnutm54IoJV4SoCdBlgRFPk26nRafnogmDq/pH7lzLI7csDvup0ABF3R6Ysji19vR4LnNIotahdOtyt5jHBnn60XxibluCy6q03Xg2bw1UNjGFN0PVrw5wpPnpB3WugEkyd4Nos5US7q80KC3ltThw1c9TYDz38rryghMnm0lPe1/ruxCMzHx9KLwZl/D3Bikxf2YLaifwphf5NIkAQiAUpaYcMamBZsABxtzE45+TBhSc4sKGJL+ZuFwXAp/gbBbxgtY+9BKHnV78v1+bcAY4Geb8kKAsUgN7Q3Tqk0mjy+L4pML2BXyD5v7x8ev9MRfKbamx1GRX+XvP1pDVLFCHsT4cnPoUoE4SjXma9s6ry0cvN2crkVfRk1KQJhbpYphmQHVfY49nNN77es0Uo+T51w5zHfvzjWwx12C1rauX5s2tdmm3mkSUNgSMsu4egE7LoGARZYqlxRlpmjzvwhDo2jKhMPnOE9lbmH9v6s3BeizMSHbtKDRgpaH1G4dqwC/rL+vIf5ED/Sfm3KEdAVadk0U8E4a3L5jqTYbSLYT1wvYCZmvvt47OKQY9ubuFkhLRht9M=
@@ -68,6 +68,7 @@ matrix:
     - rm -R node
     - tar -czf exokit-macos.tar.gz --exclude=".*" --exclude="*.tar.gz" *
     - ./exokit-codesign-macos.sh
+    - npm run test:ci
     deploy:
     - provider: releases
       api_key:
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ matrix:
     - rm -R node
     - tar -czf exokit-macos.tar.gz --exclude=".*" --exclude="*.tar.gz" *
     - ./exokit-codesign-macos.sh
-    - npm run test:ci
+    - "./node/bin/npm run test:ci"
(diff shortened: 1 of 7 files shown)


tests/unit/__setup.test.js
```diff
@@ -0,0 +1,16 @@
...
+var path = require('path');
+var Sinon = require('sinon');
+
+var core = require('../../core');
+core.setNativeBindingsModule(path.resolve('native-bindings.js'));
+
+global.assert = chai.assert;
+
+beforeEach(function () {
+  this.sinon = Sinon.createSandbox();
```
modulesio: I suspect `sinon` might fight with exokit re: some of the things we'll want to test. But that's a problem for another day.

package.json
```diff
@@ -13,7 +13,8 @@
     "install": "node ./preinstall.js && node-gyp rebuild",
     "build": "node-gyp rebuild",
     "start": "node .",
-    "test": "echo \"Error: no test specified\" && exit 0"
+    "test": "NODE_ENV=test mocha --watch tests/unit/**/*.test.js",
+    "test:ci": "NODE_ENV=test mocha tests/unit/**/*.test.js"
```
modulesio: Does this run on CI? Let's see...

I've found just booting on CI can be nontrivial because of the GL requirement.

modulesio: - https://github.com/webmixedreality/exokit/blob/master/appveyor.yml
- https://github.com/webmixedreality/exokit/blob/master/.travis.yml

ngokevin: Forgot to add `cross-env`. Pushed a fix, will see how it goes. I don't think the unit tests will invoke GL, else we can stub it out in test mode. We do the same for A-Frame.

tests/unit/__setup.test.js
ngokevin: How do you imagine it might fight?

ngokevin:
Alright, added it to Appveyor and Travis. When I had the script at `npm run test`, the test suites would always try to run it post-install...so I named them `npm run test:ci` and `npm run test:watch`.

I'm not sure the best way to set up Travis. I just added the test run in the OS X build since it seemed to set everything up.

tests/unit/__setup.test.js
modulesio: Things like timeouts, rafs, error traps, and request species are important for getting the web APIs right, and exokit has to jump a couple hoops to intercept and remap things.

If `sinon` is playing the same game I can imagine one attempting to overwrite the other. Just a hunch.

But... it would probably be fixable if we see it.

appveyor.yml
```diff
@@ -41,6 +40,7 @@ after_test:
       7z a exokit-windows-full.zip * -xr'!.git' -xr'!buildtools' -xr'!exokit-win-x64.exe'
       rm -r -fo node
       7z a exokit-windows.zip * -xr'!.git' -xr'!buildtools' -xr'!exokit-win-x64.exe' -xr'!exokit-windows-full.zip'
+      npm run test
```
modulesio: Which will this run? `test:ci` or `test:watch` or both?

.travis.yml
```diff
@@ -68,6 +68,7 @@ matrix:
     - rm -R node
     - tar -czf exokit-macos.tar.gz --exclude=".*" --exclude="*.tar.gz" *
     - ./exokit-codesign-macos.sh
+    - npm run test:ci
```
modulesio: Hm, I don't understand why this needs to be different from `appveyor.yml`.

ngokevin:
✅

.travis.yml
ngokevin: Forgot to update appveyor after I was working on Travis. They match now.

appveyor.yml
```diff
@@ -18,8 +18,7 @@ test_script:
   # Output useful info for debugging.
   - node --version
   - npm --version
-  # run tests
-  - npm test
+  - npm run test:ci
```
modulesio: I think there's no way this can run, because the actual install doesn't happen until `after_test:` (that was probably wrong for a while).

Maybe it should be renamed to `before_build`?

https://www.appveyor.com/docs/build-configuration/

ngokevin: Tried to put it in the install, but got error [`was compiled against a different Node.js version using NODE_MODULE_VERSION 64. This version of Node.js requires NODE_MODULE_VERSION 59.`](https://ci.appveyor.com/project/modulesio/exokit/build/0.0.1898).

Trying a bunch of random things, lemme know if you have an idea the proper way.

modulesio: That one in particular seems because of `nodejs_version: "9"`: https://github.com/webmixedreality/exokit/pull/153/files#diff-180360612c6b8c4ed830919bbb4dd459R5

modulesio: (We need `10` to both build and run after building)

.travis.yml
```diff
@@ -59,6 +59,7 @@ matrix:
     - export PATH="$(pwd)/node/bin:$PATH"
     - unset NVM_NODEJS_ORG_MIRROR
     - "./node/bin/npm install"
+    - "./node/bin/npm run test:ci"
```
modulesio: Yup probably a better place.

modulesio:
https://github.com/webmixedreality/exokit/commit/e4b83acbee7b4a9340641a0f433c7b13fbdacc9f passed.

- https://travis-ci.org/webmixedreality/exokit/builds/399863902
- https://ci.appveyor.com/project/modulesio/exokit/build/0.0.1921

modulesio:
✔️

PR #154, opened 2018-07-03
Add HTMLLinkElement support

modulesio:
This is mostly for fixing https://projects.gatunes.com/wsvr/#/41983/144,-144, which is a very 2D site at boot.

### DIFF
--- a/core.js
+++ b/core.js
@@ -874,6 +874,24 @@ global._runHtml = (element, window) => {
               });
           }
         }
+      } else if (el instanceof window.HTMLLinkElement) {
+        if (el.run()) {
+          if (el.childNodes.length > 0) {
+            try {
+              await _loadPromise(el)
+                .catch(err => {
+                  console.warn(err);
+                });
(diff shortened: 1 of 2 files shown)


modulesio:
This gets us to everything except drawing the chunks cubemarch. The chunks are generating since you can walk on them,.

capture 9

core.js
```diff
@@ -874,6 +874,24 @@ global._runHtml = (element, window) => {
               });
           }
         }
+      } else if (el instanceof window.HTMLLinkElement) {
+        if (el.run()) {
```
ngokevin: Perhaps this logic can be part of the class as an internal method? Also a comment of what this code does could be useful in the future.

modulesio: Agree with both.

Not actually related to this PR though, so will make it separate.

modulesio:
Tested with `node . -x none https://projects.gatunes.com/wsvr/#/41983/144,-144`, but there's still more to be done on that site.

PR #156, opened 2018-07-03
split some document-related code from core.js to src/Document.js

ngokevin:
Continuing some of the split. Later, could move Window to its own module.

### DIFF
--- a/src/Document.js
+++ b/src/Document.js
@@ -1,7 +1,190 @@
-const {HTMLElement, HTMLLoadableElement, Node} = require('./DOM');
-const symbols = require('./symbols');
+const parse5 = require('parse5');
 
-class Document extends HTMLLoadableElement {
+const DOM = require('./DOM');
+const {Event} = require('./Event');
+const GlobalContext = require('./GlobalContext');
+const symbols = require('./symbols');
+const utils = require('./utils');
+
+const _parseDocument = (s, options, window) => {
+  const ast = parse5.parse(s, {
+    sourceCodeLocationInfo: true,
+  });
+  ast.tagName = 'document';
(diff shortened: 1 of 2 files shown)


ngokevin:
Rebased.

src/Document.js
```diff
@@ -1,15 +1,199 @@
...
+        return new MouseEvent();
+      case 'Event':
+      case 'Events':
+      case 'HTMLEvents':
+        return new Event();
+      default:
+        throw new Error('invalid createEvent type: ' + type);
+    }
+  };
+  document.createRange
```
modulesio: If the point is a TODO, can we comment that?

src/Document.js
```diff
@@ -1,15 +1,199 @@
...
+  const ast = parse5.parse(s, {
+    sourceCodeLocationInfo: true,
+  });
+  ast.tagName = 'document';
+  return _parseDocumentAst(ast, options, window, true);
+};
+module.exports._parseDocument = _parseDocument;
+GlobalContext._parseDocument = _parseDocument;
+
+const _parseDocumentAst = (ast, options, window, uppercase) => {
```
modulesio: I guess that means the options arg is dead.

src/Document.js
ngokevin: Oops, accidentally got in there...

src/Document.js
ngokevin: We can pass it down later if it ends up getting used?

modulesio: I think it's attached to the `window`, which also comes in here.

I'll clean it up; not really related to this PR. 👍

PR #158, opened 2018-07-03
Abstract out remaining globalifications

modulesio:
I think just missed in https://github.com/webmixedreality/exokit/pull/151.


modulesio:
Seems to work.

PR #162, opened 2018-07-03
rebuild script

ngokevin:


package.json
```diff
@@ -12,6 +12,7 @@
   "scripts": {
     "install": "node ./preinstall.js && node-gyp rebuild",
     "build": "node-gyp rebuild",
+    "rebuild": "rm -rf node_modules && npm cache clean --force && npm install",
```
dmarcos: maybe `clean` is a better name?

dmarcos: also `rm` does not work on windows

package.json
```diff
@@ -62,6 +63,7 @@
     "puppeteer": "^1.5.0",
     "repl.history": "^0.1.4",
     "rimraf": "^2.6.2",
+    "shx": "^0.3.1",
```
modulesio: devDependencies?

modulesio:
👍

PR #166, opened 2018-07-04
fix deep cloneNode not setting parent

ngokevin:
- cloneNode was not setting `parentNode` of the recursive children it created.

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -156,21 +156,40 @@ class Node extends EventTarget {
   }
 
   cloneNode(deep = false) {
-    const el = new this.constructor();
-    el.attrs = this.attrs;
-    el.ownerDocument = this.ownerDocument;
-    el.tagName = this.tagName;
-    el.value = this.value;
-    if (deep) {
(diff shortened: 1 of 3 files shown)


src/DOM.js
```diff
@@ -471,7 +490,7 @@ class Element extends Node {
 
     this.on('attribute', (name, value) => {
       if (name === 'id') {
-        if (this.ownerDocument.defaultView[value] === undefined) {
+        if (this.ownerDocument && this.ownerDocument.defaultView[value] === undefined) {
```
modulesio: I'm not sure this is the right way to go, because having `ownerDocument` looks [non-optional](https://www.w3.org/TR/dom/#dom-node-ownerdocument) for any `Node`.

I think the reason this crashes is because of the way the test constructs the `Element`.

If we use `document.createElement` instead I think the problem goes away.

ngokevin:
Updated, thanks!

ngokevin:
Rebased, should be good to go.

modulesio:
👍

PR #167, opened 2018-07-04
only add getVRDisplays/xr to navigator if enabled

ngokevin:
three.js r93+ will check `xr in navigator` before switching over to use the WebXR manager. So the latest version of A-Frame doesn't work with Exokit because three.js WebXR doesn't work with A-Frame.

- Only add vr/xr API to navigator if enabled versus nulling it.

### DIFF
--- a/core.js
+++ b/core.js
@@ -1061,16 +1061,12 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
       result.sort((a, b) => +b.isPresenting - +a.isPresenting);
       return result;
     },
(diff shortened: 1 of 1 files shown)


modulesio:
Good catch. I wonder if it’s sensible to reverse feature detect with chrome’s origin trial tag by default?

dmarcos:
Yes, WebXR has not shippped yet on any browser. ETA is Q4 for Chrome, no dates about others. I would default to WebVR.

PR #168, opened 2018-07-04
Bugsnag only in app mode

modulesio:
`Bugsnag`+`fault-zone` hangs `mocha` tests by holding the loop. So, only add them if this is a module boot rather than a `require`.

Also, not sure if we want to be snagging test results.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,18 +12,6 @@ const url = require('url');
 const child_process = require('child_process');
 const repl = require('repl');
 
-require(path.join(__dirname, 'bugsnag'));
-require('fault-zone').registerHandler((stack, stackLen) => {
-  const message = new Buffer(stack, 0, stackLen).toString('utf8');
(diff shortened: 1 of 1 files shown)


ngokevin:
How are errors in general handled in Exokit? I often see errors crashing the browser versus being caught. I was able to start running A-Frame unit tests (which are DOM-heavy) through Karma (in-browser unit test runner) and Exokit and it'd die on test failures.

modulesio:
It's straight `node`, with all of the implications. The strategy so far has been whenever we call into user code (which is relatively well-defined), we try-catch and log. There have been cases where we weren't doing this, but if it happens it's a bug.

That is separate from this PR though. Snagging bugs should only happen when there is a full crash, not because some user handler threw. For the latter it should log and try to chug onwards best effort as any browser would.

PR #176, opened 2018-07-05
fix AudioContext.decodeAudioData being sync

ngokevin:
Was not respecting successCallback and errorCallback arguments.

Remove stub audio classes.


core.js
```diff
@@ -1960,7 +1836,21 @@ exokit.setNativeBindingsModule = nativeBindingsModule => {
...
   const {nativeAudio} = bindings;
-  AudioContext = nativeAudio.AudioContext;
+  AudioContext = class AudioContext extends nativeAudio.AudioContext {
+    /**
+     * Wrap AudioContext.DecodeAudioDataSync binding with promises and callbacks.
+     */
+    decodeAudioData(arrayBuffer, successCallback, errorCallback) {
+      try {
+        let audioBuffer = this._decodeAudioDataSync(arrayBuffer);
+        if (successCallback) { successCallback(audioBuffer); }
```
modulesio: I think this should be called async (`process.nextTick` or `setImmediate`) for consistency. There is probably user code out there that assumes that's the case.

core.js
```diff
@@ -1841,14 +1841,18 @@ exokit.setNativeBindingsModule = nativeBindingsModule => {
...
-        return Promise.resolve(audioBuffer);
-      } catch(err) {
-        if (errorCallback) { errorCallback(err); }
-        return Promise.reject(err);
-      }
+      return new Promise((resolve, reject) => {
+        process.nextTick(() => {
+          try {
+            let audioBuffer = this._decodeAudioDataSync(arrayBuffer);
+            if (successCallback) { successCallback(audioBuffer); }
```
modulesio: I would `nextTick` only the callbacks, not the whole `Promise`.

Also, this is a case where we're calling into user code from outside of a user stack, so we should `try`-`catch` and perhaps `console.warn` the catch. Otherwise, I think if the user code throws then it will be an uncaught exception and it would be snagged/bring the core down.

ngokevin:
Thanks! Passes on CI (obviously, since they're skipped), and the previously failing A-Frame test passes now.

PR #178, opened 2018-07-06
Set max listeners Infinity on EventTarget

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/177.

I think the this should cover the DOM nodes at least, since `Node` is an `EventTarget`.


modulesio:
@ngokevin is this good?

ngokevin:
Yeah, I think so.

PR #179, opened 2018-07-06
remove stubbed unbound modules, move rest of DOM classes to DOM.js (fixes #175)

### LINKED ISSUE #175: Remove unbound HTML media elements, assume native bindings.
In a thread somewhere, @modulesio said there was some unused code of unbound media elements in case we wanted to use Exokit without native bindings.

But now we drop that case and always just assume native bindings? Then we can clean up some code in core.js...moving the binding code to DOM.js, and removing the old classes from DOM.js.

ngokevin:
- When bindings are set, store in a Node module to be referenced at runtime by classes, versus using a big closure.
- Removed stubs for `ImageData`, `ImageBitmap`, `Path2D`, `CanvasGradient`, `CanvasRenderingContext2D`, `WebGLRenderingContext`, `MicrophoneMediaStream`, `HTMLImageElement`, `HTMLAudioElement`, `HTMLVideoElement`.
- Moved definitions of `HTMLImageElement`, `HTMLAudioElement`, and `HTMLVideoElement` to DOM module.
- Basic tests for img, audio, audio.


src/DOM.js
```diff
@@ -10,6 +10,7 @@ const util = require('util');
 const {defaultCanvasSize} = require('./constants');
 const {Event, EventTarget, MouseEvent} = require('./Event');
 const GlobalContext = require('./GlobalContext');
+const bindings = require('../native-bindings');
```
modulesio: Per Slack, probably don't do this. Native bindings is somewhat sensitive to when you require it, because of `Worker` threading.

https://github.com/webmixedreality/exokit/blob/0475da06f10938bdc2525ea3d8056ab6df3e2f4a/core.js#L1695

But that might be fixed when we move to Node's new `worker_threads`, which we should when it's stable.

modulesio: Then again, https://nodejs.org/api/worker_threads.html:

> Native addons are not supported yet.

Probably for similar reasons.

ngokevin:
Updated. I set the bindings in a Node module to be referenced at runtime by classes, versus waiting to define classes at runtime and using a big closure for classes to have reference to bindings.

Added some tests.

tests/unit/dom/MediaElement.test.js
```diff
@@ -0,0 +1,43 @@
...
+  var el;
+
+  beforeEach(() => {
+    document = exokit().document;
+  });
+
+  describe('<video>', () => {
+    it('can set src', () => {
+      el = document.createElement('video');
+      el.setAttribute('src', videoUri);
```
modulesio: Should probably wait for `onload`/`onerror`. Otherwise a whole class of bugs potentially fires after the test has passed.

modulesio: Sorry, for video/audio I think it's `oncanplay`.

core.js
```diff
@@ -1685,11 +1592,24 @@ exokit.setArgs = newArgs => {
...
+ *
+ * @param {string} nativeBindingsModule - Path to native bindings JS module.
+ */
 exokit.setNativeBindingsModule = nativeBindingsModule => {
   nativeBindings = true;
 
   const bindings = require(nativeBindingsModule);
 
+  // Set in binding module to be referenced from other modules.
+  for (let key in bindings) { BindingsModule[key] = bindings[key]; }
```
modulesio: Nit: could be `const`.

ngokevin: Can it? It's an iterator variable.

core.js
```diff
@@ -1685,11 +1592,24 @@ exokit.setArgs = newArgs => {
...
+  console.log(BindingsModule);
```
modulesio: Probably should keep it clean.

modulesio:
Design looks ✔️

core.js
modulesio: Pretty sure.

```
o = {lol: 'zol', root: 'toot'};
for (const k in o) {
  console.log('key', k);
}
```

The exception is `for-i` where you assign to the loop variable yourself.

ngokevin:
Updated the tests with assertion that the media gets loaded. I run a little Express server to serve files locally that will get fetched live by Exokit.

I fixed Video element not having canplay/canplaythrough, and I refactored all those handlers up to the MediaElement.

modulesio:
> I run a little Express server to serve files locally that will get fetched live by Exokit.

I think `file://` might work too. But it's good to test network since that's the main one.

tests/unit/dom/SrcableElement.test.js
```diff
@@ -0,0 +1,48 @@
+/* global assert, beforeEach, describe, it, TEST_URL */
+const fs = require('fs');
+const path = require('path');
+
+const helpers = require('../helpers');
+const exokit = require('../../../index');
+
+const audioData = fs.readFileSync(path.resolve(__dirname, '../data/test.ogg'), 'base64');
+const audioUri = `data:audio/ogg;base64,${audioData}`;
```
modulesio: This would be a good test, if it were used :).

I can do a follow-up with some more cases.

PR #182, opened 2018-07-06
Image load AsyncResource

modulesio:
_*TLDR* if you make a callback from C++ it should be wrapped in `AsyncResource`._

#### Description

We load all images asyncly on a thread. But when we were calling back to user code, we were just calling the callback and that's it.

This works, but the problem is if the user callback sets up more things on the event loop -- like a `Promise` -- then that will be queued but might not run immediately because node doesn't know about it.

That's what `AsyncResource` wrapping alleviates, so `node` knows to check the event loop for more things after the callback.

#### Why it took so long

This passed us by for so long because you usually have another thing to run soon after, and the event loop will flush when that happens. Also if we are running the loop at 90FPS then it's hard to notice any resource lag.

The problem arises when the last thing you are loading for your site is an image or something... Which is common for THREE.js. So the user sees mysterious site loading lag that will get resolved only by chance when some internal timer ticks to flush the loop.

#### Side effects

- makes some image tests run faster (900ms -> 3ms) 🏃 
- less mysterious site loading lag 🦄

Found with https://github.com/webmixedreality/exokit/pull/180.

### DIFF
--- a/deps/exokit-bindings/canvascontext/include/image-context.h
+++ b/deps/exokit-bindings/canvascontext/include/image-context.h
@@ -2,6 +2,7 @@
 #define _CANVASCONTEXT_IMAGE_H_
 
 #include <v8.h>
+#include <node.h>
 #include <nan/nan.h>
 #include <defines.h>
 #include <canvas/include/Context.h>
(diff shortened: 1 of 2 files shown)


modulesio:
@ngokevin This likely applies to places in the audio subsystem where we call back, and realtime data flow is important.

ngokevin:
Nice, I noticed the delay in the test.

I also had an issue with one of my sites where there was a really long delay after clicking the canvas. My app was doing something in a click handler causing a large delay, wonder if it's related.

Will keep eye out for other places.

PR #188, opened 2018-07-08
WebGLRenderingContext dirty flag

modulesio:
We frame-flip and VR-blit only when the [`WebGLRenderingContext` dirty flag](https://github.com/webmixedreality/exokit/blob/master/index.js#L682) is set. This is an optimization, especially applicable to contexts that do not draw every frame, like offscreen renderers.

Originally we frame-flipped only when `gl.clear()` was called.

This PR adds more cases for that: `drawElements` and `drawArrays`, and theit instanced friends.


modulesio:
This fixes some sites that do not clear but simply do a full screen redraw, such as RetroArch web player.

PR #198, opened 2018-07-08
Fix missing libopenvr_api.so

### LINKED ISSUE #197: libopenvr_api no such file crash
### System:
-Ubuntu 18.04 LTS 64bit

### Error:

```
internal/modules/cjs/loader.js:731
  return process.dlopen(module, path.toNamespacedPath(filename));
                 ^


Error: libopenvr_api.so: cannot open shared object file: No such file or directory
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:731:18)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
    at Module.require (internal/modules/cjs/loader.js:650:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/usr/local/lib/exokit/native-bindings.js:2:18)
    at Module._compile (internal/modules/cjs/loader.js:702:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
```

### Notes:
* if it crashes when (SteamVR is) not installed that would be a bug
* this one might just be an `rpath` issue with the package rather than a missing SteamVR install, but a bug nonetheless

shawwn:
Closes #197


modulesio:
Is there a solution which doesn't use `LD_LIBRARY_PATH` and/or `LD_PRELOAD`? This seems like something that should go in `binding.gyp`, as an `rpath` or a copy directive or similar.

modulesio:
Also, there's already a copy of the lib being unpacked from `native-openvr-deps` here: https://github.com/modulesio/node-native-openvr-deps

PR #209, opened 2018-07-09
Expose WebGL2 drawBuffers on root context

modulesio:
It was implemented for a while as an extension. This just exposes it on the root context per WebGL2.

https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/drawBuffers


modulesio:
It's still available via the `WEBGL_draw_buffers` extension for backwards compatibility.

PR #212, opened 2018-07-10
unit test mutation observer (fixes #183)

### LINKED ISSUE #183: Unit test MutationObserver.
I'm working on this at the moment.

I might have been getting issues with the observer missing some mutations (children getting appended), gonna unit test it to see if that's the case.

ngokevin:
- Reduce array creation, reuse empty node list.
- Base unit tests for Mutation Observer.

### DIFF
--- a/src/MutationObserver.js
+++ b/src/MutationObserver.js
@@ -1,5 +1,10 @@
+const {Node, NodeList} = require('./DOM');
+
+const emptyNodeList = new NodeList();
+
 class MutationRecord {
-  constructor(type, target, addedNodes, removedNodes, previousSibling, nextSibling, attributeName, attributeNamespace, oldValue) {
+  constructor(type, target, addedNodes, removedNodes, previousSibling, nextSibling,
+              attributeName, attributeNamespace, oldValue) {
     this.type = type;
     this.target = target;
     this.addedNodes = addedNodes;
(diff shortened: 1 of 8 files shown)


src/MutationObserver.js
```diff
@@ -138,27 +131,44 @@ class MutationObserver {
...
-      this.flush();
-    });
+    this.queue.push(new MutationRecord('attributes', el, emptyNodeList, emptyNodeList,
+                                       null, null, name, null, oldValue));
+    setImmediate(() => { this.flush(); });
   }
 
   handleChildren(el, addedNodes, removedNodes, previousSibling, nextSibling) {
-    this.queue.push(new MutationRecord('childList', el, addedNodes, removedNodes, previousSibling, nextSibling, null, null, null));
+    // Include recursive children in added and removed nodes list.
```
modulesio: I'm not sure if this is technically correct.

```
d = document.createElement('div');
d.innerHTML = `<div class=zol><div class=troll></div></div>`;
o = new MutationObserver(function(e) {
  console.log(e[0].removedNodes);
});
o.observe(d, {childList: true, subtree: true});
document.body.appendChild(d);
d.removeChild(d.childNodes[0]);
```

In Chrome you will see just the top element. In this PR you'll see both.

If this is a hack I think it's fine until some `MutationObserver` refactor, but it should be labeled as such.

src/MutationObserver.js
```diff
@@ -138,27 +131,44 @@ class MutationObserver {
...
+    if (addedNodes) {
+      for (let i = 0; i < addedNodes.length; i++) {
+        addedNodes[i].traverse(node => {
+          if (node.nodeType !== Node.ELEMENT_NODE || addedNodes.includes(node)) { return; }
```
modulesio: I think if we traverse starting at the `childNodes` then we don't need this `.includes`. I think an _actual_ DOM loop would be a serious bug.

src/MutationObserver.js
modulesio: Pre-PR Exokit behaves like Chrome (only top element in the list).

ngokevin: Hmm, I was using custom elements with the registerElement polyfill in A-Frame. If inner nodes are not reported as added, their `attachedCallback` never runs. I can double-check.

ngokevin: I will maybe move the logic to DOM innerHTML then to make sure all addedNodes are reported.

modulesio: One other approach could be https://github.com/webmixedreality/exokit/issues/183#issuecomment-403005720.

That is, we could replace the notion of `MutationObserver` [binding](https://github.com/webmixedreality/exokit/blob/master/src/MutationObserver.js#L76)[binding] and [unbinding](https://github.com/webmixedreality/exokit/blob/master/src/MutationObserver.js#L73) to every node in the tree, and just attach it to one (the observed). Then the `children`/`attribute` events can bubble up to it and can't be missed and there is nothing to keep track of other than the DOM being right.

modulesio: Also, if correctness is what we're going for (rather than a working hack), then https://github.com/webmixedreality/exokit/pull/212#discussion_r201267204 is probably a good unit test :P.

ngokevin:
Okay, so I removed those logical changes I made, and this PR is now purely unit tests (and reducing array creations). Turns out I didn't need those changes because I addressed a bug in my project.

Since everything works and passes, I'm not gonna refactor MutationObserver now, but unit tests will help in the future if anyone ever wants to.

PR #215, opened 2018-07-10
Fix args.image

shawwn:
Once `nativeWindow.swapBuffers(windowHandle);` is called, the `glReadPixels` fails to grab the frame.

This PR moves the `args.image` code to just before the `nativeWindow.swapBuffers(windowHandle);` call.


modulesio:
This seems a bit complicated to not have a writeup. Mind jotting down what we're looking at?

shawwn:
Sure. Once `nativeWindow.swapBuffers(windowHandle);` is called, the `glReadPixels` fails to grab the frame.

This PR moves the `args.image` code to just before the `nativeWindow.swapBuffers(windowHandle);` call.

PR #224, opened 2018-07-13
Capture setTimeout/setInterval

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/223

- Use integers for `requestAnimationFrame`, `setTimeout`, and `setInterval` results/cancels
- Capture `setTimeout`/`setInterval` in our core, so it ticks on `tickAnimationFrame`
- Remove dead `setTimeout`/`setInterval` when window is destroyed

### DIFF
--- a/core.js
+++ b/core.js
@@ -1473,6 +1473,8 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
   window.on('destroy', e => {
     const {window: destroyedWindow} = e;
     rafCbs = rafCbs.filter(fn => fn[symbols.windowSymbol] !== destroyedWindow);
+    timeouts = timeouts.filter(fn => fn[symbols.windowSymbol] !== destroyedWindow);
+    intervals = intervals.filter(fn => fn[symbols.windowSymbol] !== destroyedWindow);
   });
   window.history.on('popstate', (u, state) => {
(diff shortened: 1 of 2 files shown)


core.js
```diff
@@ -686,6 +687,41 @@ function tickAnimationFrame() {
 
     tickAnimationFrame.window = null;
   }
+
+  if (timeouts.length > 0) {
+    const dateNow = Date.now();
+    const removedTimeoutIndexes = {};
```
ngokevin: We can try to avoid object creations and function creations in the render loop.

e.g., also `timeouts.filter((t, i) => !removedTimeoutIndexes[i]);`, should pre-defined the function or use a reusable array that's cleared each run.

modulesio: Yeah, good call. If there's one place we should eliminate allocs it's this code path.

modulesio: Added in https://github.com/webmixedreality/exokit/pull/224/commits/639e05cbaccc17bac1e9f3a7353948918e868ca2, which should also be a general rAF optimization since we [used to create garbage in the processing](https://github.com/webmixedreality/exokit/blob/b89309146d6520d0e60a6438687578bcc8cf05ac/core.js#L658).

PR #225, opened 2018-07-13
Update CI to node 10.6.0

modulesio:
This brings in worker threads, so we can experiment with stopping using `window-worker`. That should be perf++ but more importantly stability++ since it wouldn't be a custom hack.


modulesio:
New `worker_threads` documentation from Node: https://nodejs.org/api/worker_threads.html

PR #230, opened 2018-07-16
Keep dataPath across window navigation

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/229.

The bug here was that the `dataPath` option was not threaded to `window.location` reloads, causing `localStorage` (among other things) to default to the `__dirname` for storage.

In the case of an install that's not correct, and also possibly not even writeable (such as on linux).


modulesio:
Fixed on @chrislatorres's linux box

PR #231, opened 2018-07-19
Minor tweaks to package.json

shawwn:


modulesio:
Whoops, thanks! Must have been wrong (but not causing problems) for a while.

shawwn:
It was mostly to fix the sidebar links on https://www.npmjs.com/package/exokit which currently 404.

PR #233, opened 2018-07-19
Add `npm run serve`; add renderToCanvas and xterm.js examples

shawwn:


modulesio:
Makes sense. For the examples, maybe we don't want to number them at all?

shawwn:
I'd rather rename in a separate PR.

PR #254, opened 2018-07-26
setInterval regressions

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/253.

Symptom was `setInterval` not firing at all after being set.

- Fix copy-paste errors in from `setTimeout` -> `setInterval`
- Add test suite for `requestAnimationFrame`, `setTimeout`, `setInterval`


modulesio:
Should add a regression suite for these.

PR #257, opened 2018-07-28
Add uncapped FPS command line flag

modulesio:
Adds the `-u` flag which makes Exokit not `setTimeout` to cap frames at 90 FPS.

Uncapping this cap with this flag is pointless for most displays, even XR, but it's useful for benchmarking raw performance.

With this I naively get ~400FPS in the home environment on my machine.

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,6 +44,7 @@ const args = (() => {
         'minimalFrame',
         'quit',
         'blit',
+        'uncapped',
         'require',
       ],
       string: [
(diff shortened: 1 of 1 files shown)


modulesio:
Command used for benchmarking:

```
exokit -hup # home uncapped performance-logging
```

ChrisEddy:
thanks alot! will help get benchmarks done for docs site

PR #273, opened 2018-07-30
<script async> treatment fixes

modulesio:
Previously we were not honoring the case of programmatic `document.createElement('script').async = false` followed by attachment.

In that case the script would load, but the correct behavior is to stack the scripts and execute them serially as they are attached to the `document`.

- Implement this behavior
- Add tests
- Add `script.async` accessor support


modulesio:
This is mainly to fix site loaders that rely on this browser behavior.

PR #286, opened 2018-08-01
Bugfix WebXR/WebVR render metrics

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/283.

This fixes several issues where the WebVR/WebXR/main canvas sizing was out of sync.

- The headset resolution previously assumed the base canvas context would be resized to accommodate the headset. This is _usually_ not always the case in different engines.
- Headset framebuffer was not resized on canvas framebuffer changes.


modulesio:
- [ ] Need to test on MacOS.

modulesio:
Tested and works on macOS 🎉

PR #300, opened 2018-08-03
Update README.md

Fierent:
Add new media


modulesio:
I think we went with the rolling screenshots GIF in the readme instead.

PR #316, opened 2018-08-05
Add top-level `await` for the REPL

shawwn:
```
[x] _sleep = async ms => new Promise(accept => { setTimeout(() => { accept(); }, ms); })
[AsyncFunction: _sleep]
[x] await _sleep(1000); console.log('hi'); await _sleep(1000); console.log('there');
hi
there
undefined
[x] 
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ const net = require('net');
 const child_process = require('child_process');
 const os = require('os');
 const repl = require('repl');
+const processTopLevelAwait = require('process-top-level-await');
 
 const core = require('./core.js');
 const mkdirp = require('mkdirp');
@@ -1442,7 +1443,7 @@ const _start = () => {
     const prompt = '[x] ';
 
     let lastUnderscore = window._;
-    const replEval = (cmd, context, filename, callback) => {
+    const replEval = async (cmd, context, filename, callback) => {
       cmd = cmd.slice(0, -1); // remove trailing \n
 
       let result, err = null, match;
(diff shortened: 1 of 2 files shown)


modulesio:
I'm on board philsophically, but [`process-top-level-await` seems a gigantic library](https://github.com/shawwn/process-top-level-await)...

shawwn:
You can see how it was generated here: https://github.com/shawwn/process-top-level-await/blob/98a1e1f2f726f485c7e834c27a3f435ba099993d/build.sh#L4-L8

I could forego the browserification and just export node's internal deps.

modulesio:
Doesn't that add a whole new dependency that we need to rebuild when changing the Node environment?

shawwn:
Nah, that's why I browserified it. It's completely self-contained.

modulesio:
So when node adds new APIs it will show up magically?

shawwn:
Yes. See https://github.com/nodejs/node/blob/a196aa2e24ef7069289fa4a9aa8c6862d82e6b5f/lib/internal/repl/await.js#L66 

Basically, it uses [acorn](https://www.npmjs.com/package/acorn) to parse the JS, looking for `await ...` expressions. If it finds any, it transforms the code into an equivalent `(async () => { .... })()` block. Otherwise it returns null.

modulesio:
Why not just wrap `async` always? I don't understand the need for acorn in that case.

shawwn:
It's difficult to wrap an arbitrary REPL input without acorn.

Consider how you'd transform this:

```
await _sleep(1000); console.log('hello'); await _sleep(1000); 42
```

into this:

```
(async () => {
  await _sleep(1000);
  console.log('hello');
  await _sleep(1000);
  return 42;
})();
```

(Notice the `return`.)

modulesio:
Makes sense. I'm uncomfortable with the volume of code added to make this work though; it seems like there could be corner cases, as well as incorrect/confusing stack traces with such rewriting. The complexity seems out of proportion to the benefit.

I feel like this would be better off as a Node feature than an Exokit engine feature.

PR #321, opened 2018-08-06
WebXR outputContext blitting

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/291.


modulesio:
This should now be blitting from the shared framebuffer as specified in the WebXR `requestSession`, but it's currently a black blit after the framebuffer swap.

Needs more investigation.

PR #326, opened 2018-08-07
Electron IPC for Dom2Canvas

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/325.

[WIP]


modulesio:
- [ ] If we're doing this, we should probably remove the [`puppeteer` experiment](https://github.com/webmixedreality/exokit/blob/067ddb60e1fcc79244d377707163bd5d9633e832/core.js#L36)

PR #337, opened 2018-08-08
added electron_complete.html to examples folder with mouse/keyboard event mappings

shaneharris:
Having some weird issues with the image data returned in the paint call. 

Prerequisites:

The electron browser dimensions clamp at the max screen resolution, i.e if you pass 2048x2048 on my craptop i get 1600x860.  

* The buffer doesn't seem to draw after the first draw call when using THREE.NearestFilter or if using a smaller resolution ( 1024x512 );
* The paint calls seem to return a distorted image when its just a damaged buffer. It looks like its painting the dirty rect to the wrong part of the page. 

Node: v10.8.0
OS: Windows 10 Pro x64
Exokit: 0.0.433

https://cdn.discordapp.com/attachments/474325668356947978/476851793410654209/exokit.mp4

I tried commenting out the [_flipImage](https://github.com/webmixedreality/exokit/blob/master/src/electronWorker.js#L94) method here to see if that was causing the issue but the distortion was still there. I also removed the [crop](https://github.com/webmixedreality/exokit/blob/master/src/electronWorker.js#L91) step but the distortion was still there. 

It could be an issue with three.js resizing the image as the max resolution was not POT but when i reduced the size to 1024x512 it did not paint the image ( after the first paint call ) with any of the mag/min filter settings - this is also causing the lag i assume with the scrolling.
```
THREE.WebGLRenderer: image is not power of two (1600x860). Resized to 1024x512
THREE.WebGLRenderer: image is not power of two (1600x860). Resized to 1024x512
````

### DIFF
--- a/examples/electron_complete.html
+++ b/examples/electron_complete.html
@@ -1,0 +1,325 @@
+<!doctype html>
+<html>
+<body>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/three.js/95/three.js"></script>
+<script src="https://threejs.org/examples/js/controls/OrbitControls.js"></script>
+<script>
+
+  const width = 2048;
+  const height = 2048;
+
+  let camera, scene, renderer, controls, browserPlane,
+    canvas, ctx, texture, browserWindow,
+    mouse = new THREE.Vector2(),
+    raycaster = new THREE.Raycaster();
+
+  init();
+  animate();
+
+  function init() {
+    setupScene();
+    setupCanvasAndTexture();
+    setupElectron()
+      .then(()=>{
+        setupBrowserPlane();
+        setupListeners();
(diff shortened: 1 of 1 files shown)


modulesio:
Thanks, this is an amazing example!

I got a handle on that blitting bug; secondary PR incoming...

modulesio:
Re: the POT resize for the `electron_complete.html` example, it might make sense to ditch the THREE.js texture management and use a plain `WebGLTexture` with [`texSubImage2D`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texSubImage2D).

See [here](https://github.com/webmixedreality/exokit-home/blob/master/index.html#L2538) for an example.

PR #338, opened 2018-08-09
Bugfix electron cropping

modulesio:
Fixes issues in https://github.com/webmixedreality/exokit/pull/337.

This fixes some problems with using the Electron `NativeImage` `crop` functionality in our paint transfer. Basically `crop` (done via Skia in their code) seems to mess up the stride of the resultant bitmap, so we were blitting it incorrectly into the canvas.

The fix here is to do the damage blit ourselves in JS. This is probably _more_ performant than what we were doing before with the image flip (which is still necessary), since we are flipping less pixels now.


modulesio:
capture 11

PR #341, opened 2018-08-09
Added electron test cases for resolution tests.

shaneharris:


examples/electron_tests.html
```diff
@@ -0,0 +1,442 @@
...
+    for(let i = 0;i<resolutions.length; i++) {
+
+      for (let j = 0; j < textureMagFilters.length; j++) {
+
+        for (let k = 0; k < textureMinFilters.length; k++) {
+
+          let testCount = (i+1) * (j+1) * (k+3);
+
+          await new Promise(resolve =>
+            document.addEventListener( 'keyup', () => {
```
modulesio: I think we need to remove this listener once we hit the event, or else subsequent loop iterations will trigger this, which becomes confusing.

PR #344, opened 2018-08-11
Properly initialize TEXTAREA value

shawwn:


core.js
```diff
@@ -590,6 +590,9 @@ const _fromAST = (node, window, parentNode, ownerDocument, uppercase) => {
         )
       );
     }
+    if (element.nodeName === 'TEXTAREA') {
```
modulesio: I think it makes more sense if put into an [`HTMLTextAreaElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement). Is there a technical reason not to?

shawwn: https://www.dreamsongs.com/WorseIsBetter.html

shawwn: Or rather, https://www.jwz.org/doc/worse-is-better.html

Specifically:

> Simplicity-the design must be simple, both in implementation and interface. It is more important for the implementation to be simple than the interface. Simplicity is the most important consideration in a design.

modulesio: This probably isn't the place to debate project design. For better or worse there is a standard to the rest of the code (following classy HTML specification).

If we're going to change that (which I haven't thought enough about to comment on) it should be a totally separate PR that redoes the design rather than an inconsistency introduced here.

modulesio: Hm, I don't think the delta's big enough between those two to justify the churn. I think it's fine to hold off on this PR until we can do a full `HTMLTextAreaElement` (probably only a couple methods).

modulesio:
This PR was not finished and this approach was not sufficiently general, so closing in favor of the issue: https://github.com/exokitxr/exokit/issues/1158

PR #346, opened 2018-08-11
Initial WebRTC support

### LINKED ISSUE #218: WebRTC support
This is a common feature request, since a lot of things use WebRTC for presence and chat.

The problem is the main (only?) WebRTC implementation was engineered specifically for Chrome + Firefox: https://webrtc.googlesource.com/src

It is a very large library, dabbling in many media APIs from capture to GL to codecs (has its own copy of ffmpeg for example), and just building and linking it can be a challenge (it has its own build infrastructure).

There is a node binding in the works called [`node-webrtc`](https://github.com/js-platform/node-webrtc). It includes a prebuilt bin of the main WebRTC, but it's not quite ready yet -- some things like creating media streams are missing.

https://github.com/js-platform/node-webrtc/issues/220

Still, this is something we'd like to support. Perhaps the best way forward would be to help the [`node-webrtc`](https://github.com/js-platform/node-webrtc) project complete the binding so we can slap it on `window`.

shawwn:
Closes #218

### DIFF
--- a/bugsnag.js
+++ b/bugsnag.js
@@ -21,6 +21,7 @@ const bugsnagApiKey = (() => {
 })();
 if (bugsnagApiKey) {
   bugsnag.register(bugsnagApiKey, {
+    autoNotifyUnhandledRejection: false,
     metaData: {
       argv: process.argv,
       command: GlobalContext.commands,
--- a/bugsnag.js
+++ b/bugsnag.js
@@ -9,6 +9,16 @@ const bugsnag = require('bugsnag');
 
 const GlobalContext = require('./src/GlobalContext');
 
+const IgnoredErrors = {
+  ['Failed to set local answer sdp: Called in wrong state: STATE_INPROGRESS']: true,
+  ['Failed to set local offer sdp: Called in wrong state: STATE_INPROGRESS']: true,
+  ['Failed to set remote answer sdp: Called in wrong state: STATE_INPROGRESS']: true,
+}
+
+const IgnoredPatterns = [
+  'Failed to set local offer sdp: Failed to push down transport description: Local fingerprint does not match identity.',
+];
+
 const bugsnagApiKey = (() => {
   try {
     return fs.readFileSync(path.join(__dirname, 'bugsnag.txt'), 'utf8').match(/^(\S*)/)[1];
@@ -21,7 +31,6 @@ const bugsnagApiKey = (() => {
 })();
 if (bugsnagApiKey) {
   bugsnag.register(bugsnagApiKey, {
-    autoNotifyUnhandledRejection: false,
     metaData: {
       argv: process.argv,
       command: GlobalContext.commands,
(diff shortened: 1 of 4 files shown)


bugsnag.js
```diff
@@ -21,6 +21,7 @@ const bugsnagApiKey = (() => {
 })();
 if (bugsnagApiKey) {
   bugsnag.register(bugsnagApiKey, {
+    autoNotifyUnhandledRejection: false,
```
modulesio: Does this have to do with this PR?

shawwn: Yeah, bugsnag was getting in the way of the webRTC lib. I can push a test that illustrates.

shawwn: Ok, run `npm run serve` and then `node . http://localhost:5000/webrtc`

If you remove that bugsnag line, you'll notice it exits incorrectly.

modulesio: This is a pretty drastic change though; do we know nothing else will work?

shawwn: The other thing I tried was something like this:

```
if (bugsnagApiKey) {
  bugsnag.register(bugsnagApiKey, {
    beforeSend: function (report) {
      console.log('TKTK');
      console.log(report.errorMessage);
      // Disable reporting of errors with a given error message
      if (report.errorMessage === 'Do not report this error') report.ignore()
      // or
      if (report.errorMessage === 'Do not report this error either') return false
    },
```

But `bugsnag.register` doesn't take a `beforeSend`.

If you can figure out how to get it to work with exokit's codebase (https://docs.bugsnag.com/platforms/browsers/customizing-error-reports/) then we could ignore the specific webrtc errors.

modulesio: Could you update the PR with details on why this is impossible to fix in webrtc?

At least so we know when we can remove the hack.

shawwn: Hm, which hack?

modulesio: Using external error strings (which are subject to change) as a functionality requirement.

modulesio: Hm, then it seems there is work to do on the WebRTC side understanding what's going on here...

shawwn: Maybe, but that work can probably be deferred. Google's standard webRTC examples work. https://codelabs.developers.google.com/codelabs/webrtc-web/#5

modulesio: We don’t know if it can be deferred, because we don’t understand what’s going on.

Webrtc is full of corner cases unfortunately so Its hard to gain confidence from examples.

arpu:
will this be merged soon?

modulesio:
Pinging @shawwn. There are unclear hacks in this PR which I think need to be addressed.

modulesio:
We've had WebRTC support for a while now:

- https://github.com/exokitxr/exokit/pull/765
- https://github.com/exokitxr/exokit/pull/866

Closing this PR as fixed.

PR #348, opened 2018-08-11
Fix document initialization logic

shawwn:


modulesio:
The main change here seems to be:

- If there is no `body`, then _do_ `_runHtml`
- If there is no vr displays then never `_runHtml`

Was that the intent?

src/Document.js
```diff
@@ -167,68 +167,74 @@ function initDocument (document, window) {
...
+    } else {
+      try {
+        await GlobalContext._runHtml(document, window);
+      } catch(err) {
+        console.warn(err);
+      }
 
-      document.readyState = 'interactive';
-      document.dispatchEvent(new Event('readystatechange', {target: document}));
+      document.dispatchEvent(new Event('DOMContentLoaded', {target: document}));
```
modulesio: This looks like a double `DOMContentLoaded` emit. See https://github.com/webmixedreality/exokit/pull/348/files#diff-614df7e2218ad0ff0e6a77be196d9f7cR163.

This would most likely would corrupt a site listening for this event, wouldn't it?

shawwn: image

This diff is hard to read on github, but looking at the screenshot makes it clearer.

The event is emitted only once, either in the `if` or the `else` branch. Probably worth breaking up the function into smaller pieces.

PR #352, opened 2018-08-14
apply native window size to window.innerWidth and window.innerHeight, apply pixel ratio on resize (fixes #347)

### LINKED ISSUE #347: Body clientHeight not correct
Follow-up to https://github.com/webmixedreality/exokit/pull/345#issue-207752310.

Exokit is naive about how it [computes `Element.clientHeight` right now](https://github.com/webmixedreality/exokit/blob/57ddbec6e898af1247bac2030da78a30dd698dd2/src/DOM.js#L13195). It's just the GL/canvas height, which is not correct and can be problematic if the site is reading this.

@ngokevin could you point to the repro case that should work? The one in https://github.com/webmixedreality/exokit/issues/252 is just logging `canvas`.

ngokevin:


index.js
```diff
@@ -126,10 +128,19 @@ nativeBindings.nativeGl.onconstruct = (gl, canvas) => {
...
-    const title = `Exokit ${version}`
+    const nativeWindowSize = nativeWindow.getFramebufferSize(windowHandle);
+    const nativeWindowHeight = nativeWindowSize.height;
+    const nativeWindowWidth = nativeWindowSize.width;
+
+    // Calculate devicePixelRatio.
+    window.devicePixelRatio = nativeWindowWidth / canvasWidth;
+
+    // Tell DOM how large the window is.
+    canvas.height = window.innerHeight = nativeWindowHeight / window.devicePixelRatio;
```
modulesio: Also, I think setting `canvas.{width,height}` might be doing nothing here besides wasting CPU recreating context, because the `canvas` already had this DOM metric size.

modulesio: I can fix this up.

PR #353, opened 2018-08-14
Remove canvas resize on create

modulesio:
Follow-up to https://github.com/webmixedreality/exokit/pull/352.

I don't think setting `canvas.{width,height}` is doing anything except potentially recreating context to the same size it was.

- [x] Should verify before merging


ngokevin:
Looks good to me. My test example is fine.

ngokevin:
Gonna test with updated example logging the canvas size.

ngokevin:
https://hospitable-swordfish.glitch.me/

Looks good still.

modulesio:
Thanks, didn't have to bust out a Macbook. 👍

PR #354, opened 2018-08-17
have canvas clientWidth/clientHeight read from window.innerWidth/innerHeight

ngokevin:
The canvas client size will just be the window size, right?

I was having an issue where `canvas.width` and `canvas.height` was temporarily set as the default canvas size. `getBoundingClientRect` was reading from those when I expected the current window size.

For some reason, my issue didn't reproduce in simple A-Frame examples, but it did in larger ones. Where the canvas would visible start at a certain size, and then resize itself a short time later. And at the point my application initialized, `canvas.getBoundingClientRect` was returning the default canvas size.

This fixes for me, and perhaps reduces the repetition of the pixel ratio operation, and decouples the idea of the context size versus the represented size? Don't know if it's a hack or not.

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -1862,11 +1862,11 @@ class HTMLCanvasElement extends HTMLElement {
   }
 
   get clientWidth() {
-    return this.width / this.ownerDocument.defaultView.devicePixelRatio;
+    return this.ownerDocument.defaultView.innerWidth;
   }
   set clientWidth(clientWidth) {}
   get clientHeight() {
-    return this.height / this.ownerDocument.defaultView.devicePixelRatio;
+    return this.ownerDocument.defaultView.innerHeight;
   }
   set clientHeight(clientHeight) {}
 


modulesio:
A window can have multiple `<canvas>` so window size is not necessarily any particular canvas size. If there is a hack, I think it's [this](https://github.com/webmixedreality/exokit/blob/9e20e1cb0d078aeba53a0e4fa5e0c99bfd688a72/index.js#L140). Each new canvas will "take over" the `innerWidth`.

What I _think_ might be happening here is `getBoundingClientRect` is not taking into account CSS, so whatever the window size is, it won't affect `canvas` even if styling tried to make the canvas stretchy. And this PR is kind of assuming all canvases are stretchy, so it works for that case.

It's hard to tell without a repro, but I can see how it would only appear when e.g. multiple canvases and resizes are involved.

ngokevin:
Will try to repro and look for the cause.

ngokevin:
I fixed this issue A-Frame side. Only happened on Exokit starts with a stubbed size, I needed to tell A-Frame's cursor code to update its canvas bounds when the new size came through. Didn't happen in simple examples for some reason, but will leave it at that.

modulesio:
> Wonder if it's possible to have an accurate window / canvas size from the get go and delay things until that's initialized?

It should always be accurate (to the `width`/`height` attributes of the canvas) from the start. If that's not the case, it's a bug. We default if and only if those attributes aren't provided.

However, if the `canvas` is relying on CSS layout and `getBoundingClientRect` to try to fill the `window` (i.e. not via `<canvas width height>` but with `style`) that would mean implementing heuristics for CSS layout, which is currently not done.

ngokevin:
In my case, there wasn't any CSS. The width/height is correct, just not the clientWidth/clientHeight. It started at the default canvas size, and then few seconds later, the window size would kick in. Could be an A-Frame or three.js thing...but worked around for now. https://github.com/aframevr/aframe/pull/3739

ngokevin:
There is a period where the window width and height are stubbed: https://github.com/webmixedreality/exokit/blob/master/core.js#L847 ... until the native window callbacks patches those properties.

PR #360, opened 2018-08-18
Use native node timeouts

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/357.

The problem here was twofold:

- We were taking `setTImeout`/`setInterval` into our own hands during the tick loop, so timeouts would only be ticked once per frame. That means if you do `setTimeout(fn, 0)` recursively, like some engines do, the most you'll be able to fire off is once per frame.
- `MutationObserver` was using `setImmediate` instead of `process.nextTick` to flush mutation events. Therefore events would come later than some engines expected, in terms of timeouts.

This fixes both issues; in the former case we map to the node natives while still retaining control over which `window` is doing what so we can cancel it on window destruction.


modulesio:
Here's the output from, which seems to match Chrome:

```
node .
[x] https://placid-camp.glitch.me/
undefined
[x] A-Frame Version: 0.8.2 (Date 2018-04-15, Commit #b20527f)
three Version: github:supermedium/three.js#r90fixMTLLoader
WebVR Polyfill Version: ^0.10.5
THREE.WebGLRenderer 90
This should not be <a-box>. <a-entity camera="" wasd-controls="" position="" rotation="" look-controls="" aframe-injected=""></a-entity>
<a-box> only removed after substantial timeout <a-entity camera="" wasd-controls="" position="" rotation="" look-controls="" aframe-injected=""></a-entity>
<a-box> is still not removed. <a-entity camera="" wasd-controls="" position="" rotation="" look-controls="" aframe-injected=""></a-entity>
```

ngokevin:
Looks good from my end, thanks!

PR #363, opened 2018-08-21
Get refresh rate from GLFW and use it to determine the length of a frame.

michaeltheory:


index.js
```diff
@@ -244,6 +244,8 @@ nativeBindings.nativeGl.onconstruct = (gl, canvas) => {
     })(gl.destroy);
 
     contexts.push(gl);
+    FPS = nativeWindow.getRefreshRate();
```
modulesio: This is probably right for the non-VR case, but if we enter VR we'll be capped to monitor refresh rate -- which is probably not this.

I'm thinking we'd have to do a similar thing to ask OpenVR for the refresh rate and set that instead when VR is started.

PR #366, opened 2018-08-23
Enable custom monitor selection / fullscreen for 2D canvas

michaeltheory:


core.js
```diff
@@ -206,6 +206,21 @@ class CustomElementRegistry {
 let nativeVm = GlobalContext.nativeVm = null;
 let nativeWorker = null;
 
+class MonitorManager {
+  constructor() {
+      const nativeBindingsModulePath = path.join(__dirname, 'native-bindings.js');
```
modulesio: I don't think it's a good idea to `require` this in the middle of the code.

An example of the procedure for hooking in native bindings for use is [later in this file](https://github.com/webmixedreality/exokit/blob/118928f0badefb77e1added8b06706f1ccebb2be/core.js#L1714).

PR #377, opened 2018-08-27
Support floating point textures in WebGL2

michaeltheory:


modulesio:
- [ ] Need to do this for the Magic leap/EGL case too

Although in practice I ML is doing GLES and Lumin does not have this constant, other platforms might so it's good to put the constant there.

PR #388, opened 2018-08-29
Add BUILDING.md

modulesio:


shawwn:
I think on Windows you need to install windows-build-tools.

shawwn:
To be more precise, you can avoid installing Visual Studio by installing [`windows-build-tools`](https://github.com/felixrieseberg/windows-build-tools) instead.

> Start PowerShell as Administrator and run: 

```
npm install --global --production windows-build-tools
```

Is there some reason that the full Visual Studio is required?

modulesio:
> Is there some reason that the full Visual Studio is required?

No, but that is afaik exactly what `windows-build-tools` installs, just the CLI part.

shawwn:
It installs Visual C++ Build Tools, which isn't the same thing as Visual Studio. In particular, Visual Studio ships with several CLI parts related to e.g. .NET and C# which aren't present in Visual C++ Build Tools.

modulesio:
I don't think I've ever needed `--production`

shawwn:
Still, it's in conflict with the [official instructions](https://github.com/felixrieseberg/windows-build-tools#readme). Unless there's a good reason to deviate from them, it's probably best to stick with the official installation method.

PR #395, opened 2018-08-30
Add fullscreen keybinding

vyrnix9582046173:
Fullscreen keybinding set to F11 key.

Escape key exits the fullscreen.

### DIFF
--- a/index.js
+++ b/index.js
@@ -586,6 +586,7 @@ nativeWindow.setEventHandler((type, data) => {
             handled = true;
           }
         }
+        //122 is F11 key
         if (data.keyCode === 122){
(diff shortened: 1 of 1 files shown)


chrislatorres:
Added f11 toggle

index.js
```diff
@@ -586,6 +586,15 @@ nativeWindow.setEventHandler((type, data) => {
             handled = true;
           }
         }
+        if (data.keyCode === 122){
```
modulesio: Comment that it’s F11?

chrislatorres: 2359b64 solves this

PR #399, opened 2018-08-31
Magic leap camera server (device)

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/396.


modulesio:
Browser streaming the Exokit draw app example from the ML1 device.

capture

PR #404, opened 2018-08-31
[WIP] Magic leap JS meshing API

modulesio:
The current meshing API experiment/example has a lot of room for improvement:

- It only does medium quality meshes
- It has a static mesh count limit
- It does not re-id meshes well, causing mesh updates to be missed and duplicated
- The examples never remove meshes, which is a memory leak

This PR refactors to clear up those issues, so the JS meshing API into something we can document and support.


modulesio:
Closing for https://github.com/webmixedreality/exokit/pull/415

PR #428, opened 2018-09-05
Add mixpanel

chrislatorres:
Adding [mixpanel](http://mixpanel.com) analytics to track common usage (URLs, headset used, version) and optimize. 
The `--notracking` or `--nt` flag will disable `mixpanel` and `bugsnag` tracking. 

This PR add the following mixpanel properties:
- is nativeVr used
- is nativeM for Magic Leap used
- the URL visited
- the Exokit version


index.js
```diff
@@ -1497,6 +1499,12 @@ const _prepare = () => Promise.all([
 
 const _start = () => {
   let {url: u} = args;
+  mixpanel.track('Exokit Start', {
+     MagicLeap: nativeMl != null,
```
modulesio: Maybe just `!!` for both?

index.js
```diff
@@ -3,6 +3,8 @@
 const cwd = process.cwd();
 process.chdir(__dirname); // needed for global bin to find libraries
 
+const Mixpanel = require('mixpanel');
+const mixpanel = Mixpanel.init('7c1cbc95d6519b7d6417d6fe1ce393e3');
```
modulesio: Probably would:
- break this code out to a text file
- require it much lower in the booting, so it's not loaded in the module case and close to bugsnag
- have a flag to disable it

index.js
chrislatorres: fixed by bbdd6a9

index.js
chrislatorres: flag added at ff8fdaf
moved require at  3f427dd
load key at 608d3ba

index.js
```diff
@@ -1609,6 +1622,9 @@ const _start = () => {
 if (require.main === module) {
   if (nativeAnalytics) {
```
modulesio: Maybe this whole block should be under the `nt` flag as well?

index.js
```diff
@@ -17,9 +17,10 @@ const core = require('./core.js');
 const mkdirp = require('mkdirp');
 const replHistory = require('repl.history');
 const minimist = require('minimist');
+const ip = require('ip');
+const {version} = require('./package.json');
```
modulesio: Nit, but I would leave `package.json` below. It's not a module to be included, so probably doesn't belong with those.

index.js
```diff
@@ -1497,6 +1501,15 @@ const _prepare = () => Promise.all([
 
 const _start = () => {
   let {url: u} = args;
+  if (!args.notracking) {
```
modulesio: Maybe also `!nativeAnalytics` here, or else platforms like magic leap (no `nativeAnalytics`) might crash.

mpkey.txt
```diff
@@ -0,0 +1 @@
+7c1cbc95d6519b7d6417d6fe1ce393e3
```
ngokevin: Might be more clean for this to be left out and only built into the production version of Exokit?

ngokevin: This can be a JSON.

modulesio: > Might be more clean for this to be left out and only built into the production version of Exokit?

Yeah, that's probably the route to take. For dev + forks we don't want to be doing analytics.

> This can be a JSON.

Indeed 👍.

chrislatorres:
Closing this issue due to being outdated.

PR #441, opened 2018-09-09
make sure _normalize converts to original promise prototype in case was overwritten w/ polyfill (fixes #433)

### LINKED ISSUE #433: Issues with core-js es6 Promise polyfill.
Test case: https://webvr.soundboxing.co/

Firefox and Exokit both get routed to use the es6 Promise polyfill, commonly used in Babel projects. There are issues with `state.reactions.push(reaction)` (or `this._c.push`) where `state.reactions` (or `this._c`) is undefined in Exokit, whereas it should be an array defined in the internal Promise constructor. Works in Firefox.

Maybe the native Promise objects is clashing with the polyfill Promise somewhere, or something to do with sandboxing. Looking into it.

ngokevin:
Approach was to create and store a dummy window object which will can reference the original types from. I suppose we could also update vm-one to expose the v8::Promise protoype too?

Test case: webvr.soundboxing.co


src/utils.js
```diff
@@ -71,36 +71,71 @@ const _runJavascript = (jsString, window, filename = 'script', lineOffset = 0, c
...
+ * Or convert JS object when passing to native.
+ * Node's vm module was slow so there's vm-one for one-way bindings.
+ * github.com/modulesio/vm-one
+ * nativeVm.setPrototype will do the nan casting setPrototype between types.
+ * Required for instanceof to work.
+ *
+ * @param obj - Object to convert.
+ * @param targetContext - window or vm context that prototype will get retrieved from.
+ */
+const _normalizeBuffer = (function () {
```
modulesio: `_normalizeBuffer` is probably a misnomer. It's more like `_convertNativePrototype`.

src/utils.js
```diff
@@ -71,36 +71,71 @@ const _runJavascript = (jsString, window, filename = 'script', lineOffset = 0, c
...
+      dummyWindow = vmo.getGlobal();
+      dummyWindow.vm = vmo;
     }
-    case 'Blob': {
-      if (!(b.buffer instanceof target.Buffer)) {
-        GlobalContext.nativeVm.setPrototype(b.buffer, target.Buffer.prototype);
+
+    const name = obj && obj.constructor && obj.constructor.name;
+
+    switch (name) {
```
modulesio: If we go the above route, maybe this whole constructor name checking is pointless, since we can check actual prototypes instead, which will be more accurate.

src/utils.js
```diff
@@ -71,36 +71,71 @@ const _runJavascript = (jsString, window, filename = 'script', lineOffset = 0, c
...
+      }
+      case 'Blob': {
+        if (!(obj.buffer instanceof targetContext.Buffer)) {
+          GlobalContext.nativeVm.setPrototype(obj.buffer, targetContext.Buffer.prototype);
+        }
+        break;
+      }
+      case 'Promise': {
+        if (!(obj instanceof targetContext[name])) {
+          if (targetContext.prototype === dummyWindow.prototype) {
```
modulesio: This fix could probably be used for all of the other types too, not just `Promise`. I suspect there are libs out there that would make other classes suffer this same problem when the global is overwritten.

ngokevin:
Addressed.

src/utils.js
```diff
@@ -72,71 +73,61 @@ const _runJavascript = (jsString, window, filename = 'script', lineOffset = 0, c
...
-    if (!dummyWindow) {
-      let vmo = GlobalContext.nativeVm.make();
-      dummyWindow = vmo.getGlobal();
-      dummyWindow.vm = vmo;
-    }
+  const isToWindow = !!targetContext[symbols.prototypesSymbol];
 
-    const name = obj && obj.constructor && obj.constructor.name;
+  // Use Buffer for Blob.
+  if (name === 'Blob') { name = 'Buffer'; }
```
modulesio: Is this the same thing the previous code was doing?

`Blob` is not itself a `Buffer`, but it does have a `Buffer` under `Blob.buffer` -- _and_ that is what needs a prototype conversion, not the `Blob`.

modulesio: Previous code: https://github.com/webmixedreality/exokit/pull/441/files#diff-2b4ca49d4bb0a774c4d4c1672d7aa781L96

src/utils.js
```diff
@@ -71,37 +72,65 @@ const _runJavascript = (jsString, window, filename = 'script', lineOffset = 0, c
...
+
+  // Normalize to window prototype.
+  if (isToWindow) {
+    GlobalContext.nativeVm.setPrototype(obj, targetContext[symbols.prototypesSymbol][name] ||
+                                             targetContext[name].prototype);
+    return obj;
+  }
+
+  // Normalize to native prototype.
+  GlobalContext.nativeVm.setPrototype(obj, targetContext[name].prototype);
```
modulesio: Seems we can fall though here if `obj` is something unexpected, like `null`. I don't know if trying to set the prototype will explode in that case, but I _do_ think callers are expecting this function to be safe for all inputs.

The original code was fine since it wouldn't [pass the `name` check](https://github.com/webmixedreality/exokit/pull/441/files#diff-2b4ca49d4bb0a774c4d4c1672d7aa781L76).

We should either confirm that this is fine for all inputs (hard) or be safe and do e.g. as prelude to this function

```
if (obj && typeof obj === 'object') { // non-null object
```

We also would probably be guaranteed to have a `constructor.name` in that case.

ngokevin: Done, added unit tests.

src/utils.js
```diff
@@ -96,7 +122,7 @@ const _normalizePrototype = (obj, targetContext) => {
 
   // Normalize to window prototype.
   if (isToWindow) {
-    GlobalContext.nativeVm.setPrototype(obj, targetContext[symbols.prototypesSymbol][name] ||
+    GlobalContext.nativeVm.setPrototype(obj, targetContext[symbols.prototypesSymbol][name].protoype ||
```
modulesio: `prototype`

ngokevin: whoops, fixed

PR #445, opened 2018-09-09
remove keyboard image from readme

ngokevin:
Was kind of weird there's just a random keyboard in the readme. Still in the assets folder though.


modulesio:
Yeah, it probably doesn't belong there. This is for the WebXR emulation/reality tabs but that should go elsewhere.

PR #446, opened 2018-09-09
source three.js from npm

ngokevin:


modulesio:
I think this is fine, but is there some impetus for the full build? I think the only things that are being used internally are the maths.

ngokevin:
Oh, didn't realize it was just the math. Cleans the repo a bit. I'm fine either way, close or keep.

PR #459, opened 2018-09-12
Src/scripts/metadata files cleanup

modulesio:
File restructuring continued from https://github.com/webmixedreality/exokit/pull/430.


ngokevin:
thanks for the cleanups

PR #462, opened 2018-09-13
Renderloop dirty GL

modulesio:
This fixes an issue where not dirtying the `<canvas>` WebGL context would cause the frame to not be submitted to the AR/VR device (especially magic leap). In some cases this is not tolerated well -- i.e. if you do not end a frame that you started the native loop will fail to render anymore.

- This changes it so that any time you have a dirty, visible context, it will go through the swapchain.
- Additionally, the Magic Leap case to always be considered dirty for swap purposes.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -358,6 +358,6 @@ int main(int argc, char **argv) {
 
     ML_LOG_TAG(Info, LOG_TAG, "---------------------exokit end");
 
-      return 0;
+    return 0;
   }
 }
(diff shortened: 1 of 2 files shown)


ngokevin:
I started getting big regular framedrops with this when using with I-Frames.

screen shot 2018-09-14 at 3 38 54 pm

screen shot 2018-09-14 at 3 39 02 pm

screen shot 2018-09-14 at 3 39 08 pm

modulesio:
Thanks, it's probably the [switching window context](https://github.com/webmixedreality/exokit/pull/462/files#diff-1fdf421c05c1140f6d71444ea2b27638R904) and [`flush`](https://github.com/webmixedreality/exokit/pull/462/files#diff-1fdf421c05c1140f6d71444ea2b27638R905).

PR #468, opened 2018-09-14
always flush when dirty in _blit and set window context

ngokevin:
The conditionals were a bit tangled. We should always set the current window context and flush it if it is dirty. Before, if it was dirty and not visible, it wouldn't grab the current window handle but stuck on the previous loop iteration's one when it passed onto `nativeWindow.swapBuffers(windowHandle);`.

https://github.com/webmixedreality/exokit/pull/462#issuecomment-421502405


ngokevin:
Yup!

PR #469, opened 2018-09-14
Update ndb instructions for require path.

ngokevin:


modulesio:
Missed in https://github.com/webmixedreality/exokit/pull/459.

PR #470, opened 2018-09-15
make window.Promise read-only and not overwritable by polyfills

ngokevin:
The `normalizePrototype` fix from earlier did get us pass an explicit error, but now we have unexpected behavior from the Promise object being force-polyfilled by some sites.

With webvr.soundboxing.co using babel-polyfill, A-Frame did not like the Promise being overwritten. 
`Promise.all` was not working as expected, stopping the scene from loading. Perhaps to make it easier, don't allow Promise to be overwritten by polyfills that aren't needed.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1540,6 +1540,18 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
     }
   });
 
+  // Block polyfills from overwriting Promise object.
+  const originalWindowPromise = window.Promise;
+  Object.defineProperty(window, 'Promise', {
+    get() {
+      return originalWindowPromise;
+    },
+    set() {
+      console.warn('window.Promise cannot be overwritten.');
+      return;
+    }
+  });
+
   if (!parent) {
     window.tickAnimationFrame = tickAnimationFrame;
 
--- a/tests/unit/Window.test.js
+++ b/tests/unit/Window.test.js
@@ -1,0 +1,11 @@
+/* global assert, describe, it */
+const exokit = require('../../src/index');
+
+describe('Promise', () => {
+  it('cannot be overwritten', () => {
+    const window = exokit().window;
+    const originalPromise = window.Promise;
+    window.Promise = () => {};
+    assert.equal(window.Promise, originalPromise);
+  });
+});


src/core.js
```diff
@@ -1540,6 +1540,18 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
     }
   });
 
+  // Block polyfills from overwriting Promise object.
+  const originalWindowPromise = window.Promise;
+  Object.defineProperty(window, 'Promise', {
```
modulesio: Is `writable`/`enumerable` more appropritate? Probably makes more sense than a function call trap.

```
$ node
x = {};
Object.defineProperty(x, 'lol', { value: 'zol', writable: false, enumerable: true, });
x.lol = 'woot';
x.lol;
```

ngokevin:
I can try to keep finding the underlying bug, I don't think it's prototype conversion since all the `Promise.all`s are from within the site, and `Promise.all` is a normal function that returns a Promise.

modulesio:
Is there a smaller repro for this?

ngokevin:
Yeah: https://glitch.com/edit/#!/leaf-anchovy?path=index.html:1:0

```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>Hello, WebVR! • A-Frame</title>
    <meta name="description" content="Hello, WebVR! • A-Frame">
    <script src="https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/7.0.0/polyfill.js"></script>
    <script src="https://aframe.io/releases/0.8.2/aframe.min.js"></script>
    <script>
      var promise = new Promise(resolve => {
        setTimeout(() => { resolve(); }, 1000);  
      });
      
      Promise.all([promise]).then(() => { console.log('Promise.all resolved.'); });
    </script>
  </head>
  <body>
    <a-scene background="color: #ECECEC">
      <a-box position="-1 0.5 -3" rotation="0 45 0" color="#4CC3D9" shadow></a-box>
      <a-sphere position="0 1.25 -5" radius="1.25" color="#EF2D5E" shadow></a-sphere>
      <a-cylinder position="1 0.75 -3" radius="0.5" height="1.5" color="#FFC65D" shadow></a-cylinder>
      <a-plane position="0 0 -4" rotation="-90 0 0" width="4" height="4" color="#7BC8A4" shadow></a-plane>
    </a-scene>
  </body>
</html>
```

ngokevin:
Both the promise and Promise.all are instances of the polyfilled Promise, so it's not like a mismatch or anything. Maybe it's the polyfill, something in its JS doesn't work well with Exokit.

https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/7.0.0/polyfill.js

modulesio:
Thanks, I could probably debug with the above.

With things like this it could betray a different more fundamental issue, such as async resource callbacks not inducing libUV to tick, so I'd rather understand what's happening than forcefully override it.

This is probably a case where `console.log` needs to be analysed to understand the behavior difference in this polyfill.

ngokevin:
Will move to an issue.

PR #471, opened 2018-09-15
Events system DOMification

modulesio:
The existing way of doing event management for DOM events was a bit hacky. Every `EventTarget` was basically a `node` `EventEmitter`.

The problem with doing this is that some frameworks (notably A-Frame) have their own event system and have naming clashes with `node`-style `EventEmitter`. That is not the DOM API so Exokit was at fault here, necessitating renaming `emit` to `_emit` and other fragile weirdness.

This PR re-implements events in term of `addEventListener` with no `node`-style `EventEmitter`, which makes the code more readable and less fragile. This came about as a side-effect of debugging DOM event emits.

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -1819,7 +1819,8 @@ class HTMLIFrameElement extends HTMLSrcableElement {
             }
           })
           .catch(err => {
-            console.error(err);
+            console.warn(err);
+
             this.dispatchEvent(new Event('load', {target: this}));
           })
           .finally(() => {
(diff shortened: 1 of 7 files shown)


modulesio:
- [ ] Need to fix tests

modulesio:
Might actually not want to do this, for perf reasons.

I was mistaken when I thought this was a bug manifestation but I think it was other DOM bugs we recently fixed.

modulesio:
Going to table this for now.

Issue #361, opened 2018-08-18
Mutation Observer should not bubble up `children` events to document from non-attached elements.


ngokevin:
In A-Frame, the child's `attachedCallback` is getting called when appending an element to an unattached element.

Reproduction case: https://coconut-mass.glitch.me/

I can try to work on this. Should be walking up the DOM to see if it's attached to the document or not, and tweaking the Mutation Observer related event emissions as appropriate.

ngokevin:
Don't fully understand #219. Still trying to track down, cross-referencing with the [custom elements polyfill](https://github.com/dmarcos/document-register-element/blob/master/src/document-register-element.js).

A simplified reproduction case: https://flint-language.glitch.me/

PR #481, opened 2018-09-16
[MutationObserver] only check if mutated el is part of document if observing document (fixes #473, continuation #361)

### LINKED ISSUE #473: Unexpected behavior with Promise.all when Promise is polyfilled by babel-polyfill/core-js.
https://github.com/webmixedreality/exokit/pull/470#issuecomment-421528903

https://glitch.com/edit/#!/leaf-anchovy?path=index.html:1:0

Promise.all never resolves.

```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>Hello, WebVR! • A-Frame</title>
    <meta name="description" content="Hello, WebVR! • A-Frame">
    <script src="https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/7.0.0/polyfill.js"></script>
    <script src="https://aframe.io/releases/0.8.2/aframe.min.js"></script>
    <script>
      var promise = new Promise(resolve => {
        setTimeout(() => { resolve(); }, 1000);  
      });
      
      Promise.all([promise]).then(() => { console.log('Promise.all resolved.'); });
    </script>
  </head>
  <body>
    <a-scene background="color: #ECECEC">
      <a-box position="-1 0.5 -3" rotation="0 45 0" color="#4CC3D9" shadow></a-box>
      <a-sphere position="0 1.25 -5" radius="1.25" color="#EF2D5E" shadow></a-sphere>
      <a-cylinder position="1 0.75 -3" radius="0.5" height="1.5" color="#FFC65D" shadow></a-cylinder>
      <a-plane position="0 0 -4" rotation="-90 0 0" width="4" height="4" color="#7BC8A4" shadow></a-plane>
    </a-scene>
  </body>
</html>
```

> With things like this it could betray a different more fundamental issue, such as async resource callbacks not inducing libUV to tick, so I'd rather understand what's happening than forcefully override it.

Source of the polyfill: https://raw.githubusercontent.com/zloirock/core-js/master/packages/core-js/modules/es.promise.js

I see it has some async calls in this microtask library: https://raw.githubusercontent.com/zloirock/core-js/master/packages/core-js/internals/microtask.js and https://raw.githubusercontent.com/zloirock/core-js/master/packages/core-js/internals/task.js

ngokevin:


tests/unit/__setup.test.js
```diff
@@ -7,7 +7,7 @@ var Sinon = require('sinon');
 const PORT = 10000;
 
 var core = require('../../src/core');
-core.setNativeBindingsModule(path.resolve('src', 'native-bindings.js'));
+core.setNativeBindingsModule('../src/native-bindings.js');
```
modulesio: I see the `..` bug, but I think for windows using `path.resolve()` is important since it gives the absolute `C:\`-style path.

ngokevin: Updated to resolve off __dirname.

PR #483, opened 2018-09-16
update exokit.sh path

chrislatorres:
Updating paths to support changes to file locations.


chrislatorres:
at https://github.com/webmixedreality/exokit/blob/master/tests/unit/__setup.test.js#L10

in exokit/tests/unit running node ../../src/native-bindings.js has the error for libopenvr_api.so
but in exokit/tests
running node ../src/native-bindings.js runs fine

src/native-bindings.js
```diff
@@ -1,5 +1,5 @@
 const path = require('path');
-const bindings = require(path.join(__dirname, '../', 'build', 'Release', 'exokit.node'));
+const bindings = require(path.join(__dirname, '..', 'build', 'Release', 'exokit.node'));
```
modulesio: Why this change? Does it do anything?

chrislatorres: It doesn't do anything technically, it makes it less confusing without the slash.

PR #487, opened 2018-09-16
add various webgl extensions

ngokevin:
Adding various webgl extensions. The remaining ones are:

- EXT_disjoint_timer_query
- WEBGL_compressed_texture_astc
- WEBGL_compressed_texture_atc
- WEBGL_compressed_texture_etc
- WEBGL_compressed_texture_s3tc_srgb
- WEBGL_debug_shaders
- WEBGL_lose_context

But going to move on for now.

### DIFF
--- a/src/native-bindings.js
+++ b/src/native-bindings.js
@@ -61,7 +61,7 @@ bindings.nativeGl = (nativeGl => {
   }
   WebGLRenderingContext.onconstruct = null;
   return WebGLRenderingContext;
-})(bindings.nativeGl);
+})(bindings._nativeGl);
 bindings.nativeGl2 = (nativeGl2 => {
   function WebGL2RenderingContext(canvas) {
     const gl = new nativeGl2();
@@ -77,7 +77,7 @@ bindings.nativeGl2 = (nativeGl2 => {
     WebGL2RenderingContext[k] = nativeGl2[k];
   }
   return WebGL2RenderingContext;
-})(bindings.nativeGl2);
+})(bindings._nativeGl2);
 
 const {PannerNode} = nativeAudio;
 PannerNode.setPath(path.join(require.resolve('native-audio-deps').slice(0, -'index.js'.length), 'assets', 'hrtf'));
--- a/deps/exokit-bindings/webglcontext/include/webgl.h
+++ b/deps/exokit-bindings/webglcontext/include/webgl.h
@@ -221,12 +221,12 @@ public:
   static NAN_METHOD(VertexAttrib2fv);
   static NAN_METHOD(VertexAttrib3fv);
   static NAN_METHOD(VertexAttrib4fv);
-  
+
   static NAN_METHOD(VertexAttribI4i);
   static NAN_METHOD(VertexAttribI4iv);
   static NAN_METHOD(VertexAttribI4ui);
   static NAN_METHOD(VertexAttribI4uiv);
-  
+
   static NAN_METHOD(VertexAttribDivisor);
   static NAN_METHOD(VertexAttribDivisorANGLE);
   static NAN_METHOD(DrawBuffers);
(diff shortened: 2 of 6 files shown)


modulesio:
Is the implication that there is process memory retained across executions?

ngokevin:
Found the issue sort of. Seems like the bindings module was cached somehow, but native-bindings would run again on subsequent test runs. bindings.nativeGl(2) would get re-wrapped again. Made it so it isn't overwriting the original bindings.nativeGl(2) and that works.

src/native-bindings.js
```diff
@@ -61,7 +61,7 @@ bindings.nativeGl = (nativeGl => {
   }
   WebGLRenderingContext.onconstruct = null;
   return WebGLRenderingContext;
-})(bindings.nativeGl);
+})(bindings._nativeGl);
```
modulesio: I don't quite understand the need for this. Was this code running twice in one execution or something?

ngokevin: bindings.js only gets required once and kept throughout test executions. But all the other modules will re-run, causing bindings.js to get overwritten and rewrapped.

There's a comment in bindings.js: `Due to the way `require` works with multithreading (used for `Worker`), we use this route to make sure the binaries only get initialized once. This might change with node's native "worker" support that's coming.`

modulesio: > the other modules will re-run

Which other modules is this referring to?

That comment probably doesn't apply to `bindings.js`, but `setNativeBindingsModule`.

#### Overview as I see it

- `setNativeBindingsModule` is called on bootstrap
- `setNativeBindingsModule` requires `native-bindings.js` which requires the node binary once, and then wraps it in JS
- `setNativeBindingsModule` attaches the result of this to `bindings.js` for global exposure, but this is a temporary hack in file renaming
 - the above should happen once on startup, and bindings should never be touched again, either reassigned or changed; if they are that seems like possibly a bug

modulesio: Would the [tests call](https://github.com/webmixedreality/exokit/blob/master/tests/unit/__setup.test.js#L10) be causing problems?

I think if you require `exokit`'s main, you'll run `setNativeBindingsModule`. But tests are running that as well.

ngokevin: I removed the call from setNativeBindingsModule which is harmless, but if I revert the change that prevents duplicate wrapping of nativeGl, then the test still fails. 

Perhaps it's still better to make native-bindings.js more robust in case it gets invoked more than once, else it's pretty fragile. 

> the above should happen once on startup, and bindings should never be touched again, either reassigned or changed; if they are that seems like possibly a bug

I guess it happens again due to the way the test runner works in watch mode, re-requiring everything and re-initializing *all* the modules in the codebase. I'm not 100% sure what's happening under the hood, if you want to poke around. If I updated the tests in watch mode 10 times, the nativeGl wrap would get wrapped 10 times; I just assumed `bindings.js` stayed static while `native-bindings.js` got re-initialized multiple times.

PR is ready otherwise though.

modulesio: > I guess it happens again due to the way the test runner works in watch mode, re-requiring everything and re-initializing all the modules in the codebase.

I see, my hunch is that's probably not going to work, especially once `Worker` gets involved in other threads.

Basically if the `_nativeGl` my theory is it's covering up a much more serious problem, such as `mocha` watch not being compatible with the threading architecture.

I'll take the PR for a spin.

modulesio:
Continued in https://github.com/webmixedreality/exokit/pull/499.

src/native-bindings.js
modulesio: I ended up solving this without and rename or chance of re-wrapping by making sure to create a fresh object on re-require: https://github.com/webmixedreality/exokit/pull/499/commits/0211d47e7a67c655c2e26ada0944edba5a6094e3

PR #488, opened 2018-09-17
support relative path hrefs in HTMLAnchorElement

ngokevin:


src/DOM.js
```diff
@@ -1209,6 +1209,10 @@ class HTMLElement extends Element {
 }
 module.exports.HTMLElement = HTMLElement;
 
+function getAnchorUrl(anchorEl) {
+  return new url.URL(anchorEl.href, anchorEl.ownerDocument.defaultView.location.origin);
```
modulesio: Should this be `href` or `pathname` instead?

If you try on github:

```
a = document.createElement('a'); a.href = 'lol.txt'; a.pathname;
```

"/webmixedreality/exokit/pull/488/lol.txt"

ngokevin: Updated.

PR #495, opened 2018-09-18
Clean up attribute DOM.js src empty checks

modulesio:
Cleanup/code reduction from https://github.com/webmixedreality/exokit/pull/491.


ngokevin:
looks good

PR #496, opened 2018-09-19
ML camera livestream

modulesio:
This adds a fast camera stream renderer + compositor + streamer for Magic Leap One. This allows opening the full camera view as a regular web page from the device.

This functionality already existed as a hack via the screenshot API, but this is a fully realtime implementation using the camera preview stream.

Included is a frontend example that consumes the jpeg image stream to render it.


machenmusik:
Is this just video and no audio?

modulesio:
Correct, this PR at least has nothing to do with audio.

PR #403, opened 2018-08-31
Android (ARCore) support

chrislatorres:
This PR will add Android ARCore functionality to Exokit.

PR #514, opened 2018-10-01
fix webgl.h includes for android

ngokevin:
- Move the Android GLES includes up because Android compiler sets `__linux__` wouldn't have included the right libraries.
- Include Android also in using egl over glfw.


chrislatorres:
In case this is useful, began attempt to piggyback Android build off ML at #403

ngokevin:
Cool, I'll try out the build. I'm looking at it for Oculus support, not ARCore, so it shouldn't go the ML route past the build/compilation.

PR #533, opened 2018-10-15
Non c drive

machenmusik:
(Note that there seem to be other C drive dependencies, so my D drive build still fails at the moment)


modulesio:
Even if it’s not 100% this looks like progress in the right direction.

machenmusik:
looks like in combination with the magicleap-js change, it will work.

PR #534, opened 2018-10-15
Continue gen-dlibs-h script if unmatched

machenmusik:
```
avaer [8:27 PM]
Ok now you can probably skip the non matching ones here
In gen-dlibs
```


modulesio:
Thanks, picked. 👌

Issue #551, opened 2018-10-17
Move exokit.{sh,cmd} to root


modulesio:
Right now these are in `scripts/` but the expectation is the main EXE should probably be in root.

We should move the EXEs to root and patch the launcher paths to see it there.

chrisplatorres:
Is this still a relevant change?

modulesio:
No, replaced by the launcher. Closing per https://github.com/exokitxr/exokit/pull/552#issuecomment-487256401

PR #552, opened 2018-10-17
Update script paths

chrisplatorres:
This PR is to address #551 :

- move `scripts/exokit.sh` and `scripts/exokit.cmd` to root directory 
- update paths


modulesio:
Are the macOS/windows installers still good after this?

chrisplatorres:
Not sure if moving the scripts to root still makes sense, as there is now going to be a move to using the launcher.

modulesio:
Yeah, I think this requirement would be subsumed by the launcher. The confusion of script location is probably much better solved that way.

Administratively closing for that reason.

PR #562, opened 2018-10-18
Update debug-ml.sh command entrypoint

modulesio:
Fixes `gdb` command not found errors.

(thanks @machenmusik)


modulesio:
This mostly affects `python` lookup.

Issue #537, opened 2018-10-16
ML: audio context does not survive destroy


machenmusik:
https://speech-polyfill.azurewebsites.net/

While verifying #484 using #477,
some debugging indicates that things are working until recognition.stop() is called,
which tries to destroy the audio context, as a new one will be created upon recognition.start()

On ML (did not test others), recognition.stop() does not successfully return, and thus recognition becomes and remains broken.

machenmusik:
from Slack a while back:


looks like recognition.stop() doesn't come back

machenmusik [9:44 AM]
... looks like that destroys audio context https://github.com/Azure-Samples/SpeechToText-WebSockets-Javascript/blob/master/src/common.browser/MicAudioSource.ts#L159
src/common.browser/MicAudioSource.ts:159
    ```public TurnOff = (): Promise<boolean> => {```
Azure-Samples/SpeechToText-WebSockets-JavascriptAdded by GitHub
and then a new one is created when recognition is restarted https://github.com/Azure-Samples/SpeechToText-WebSockets-Javascript/blob/master/src/common.browser/MicAudioSource.ts#L55
src/common.browser/MicAudioSource.ts:55
    ```public TurnOn = (): Promise<boolean> => {```
Azure-Samples/SpeechToText-WebSockets-JavascriptAdded by GitHub
... and exokit doesn't survive the TurnOff attempt

machenmusik:
0.0.494 still has the issue.

modulesio:
Whoops thanks, didn’t mean to close this yet.

modulesio:
Added the speech demo link to the OP.

modulesio:
Writing down for posterity:

The main hang we have now is that the audio context destruction does not return.

After analysis this is the result of context destruction is:

- waiting for the context to finish rendering requires joining the Labsound audio graph thread
- the audio graph thread has a tail time to account for flushing existing buffers, and waits for enough sample time to pass through the render thread
- time only passes in the render thread while rendering happens
- if there is a `ScriptProcessorNode` in the render chain, that requires calling back to the main thread to execute the user logic
- we cannot call back to the main thread because the main thread is waiting for the context to finish rendering in the first place
 - this is a deadlock

Seems the right solution would be to make the tail unwind asynchronous.

modulesio:
[...]

Issue #477, opened 2018-09-15
Trying to use Web Speech API polyfill doesn't work.


machenmusik:
https://speech-polyfill.azurewebsites.net/

I understand that ExoKit currently doesn't implement the Web Speech API, so tried using a polyfill that uses getUserMedia() to get the audio, which works on Firefox - https://github.com/anteloe/speech-polyfill - but that doesn't seem to work with ExoKit.

modulesio:
It looks like it uses `WebRTC`.

That would make this blocked on https://github.com/webmixedreality/exokit/issues/218.

machenmusik:
Looks to me as though it polyfills from old to new gUM API, and uses `microsoft-speech-browser-sdk` which uses `getUserMedia` not all of WebRTC... 

```Browser support
The SDK depends on WebRTC APIs to get access to the microphone and read the audio stream. Most of todays browsers(Edge/Chrome/Firefox) support this. For more details about supported browsers refer to navigator.getUserMedia#BrowserCompatibility

Note: The SDK currently depends on navigator.getUserMedia API. However this API is in process of being dropped as browsers are moving towards newer MediaDevices.getUserMedia instead. The SDK will add support to the newer API soon.```

How does ExoKit currently expose the microphone, is there a simple sample page?
I thought it was providing `navigator.mediaDevices.getUserMedia` for audio, but not for video...

modulesio:
`getUserMedia` is a separate from `WebRTC`. You can put a media stream into `WebRTC`, but that's not the same thing. That speech service appears to require `WebRTC`.

Exokit exposes `navigator.mediaDevices.getUserMedia`, for video and audio and this is not related to `WebRTC`. But WebRTC is on the roadmap since there is demand for that despite the technical problems.

The home environment has microphone support: https://github.com/webmixedreality/exokit-home/blob/7f1247b45cbf98d370b0a6b3228617c7a3b0ef9f/index.html#L381

And there are video/webcam examples in this repo as well: https://github.com/webmixedreality/exokit/blob/master/examples/webcam.html

It would be nice to get a microphone example in this repo though.

machenmusik:
I guess I am not understanding where you think the underlying Microsoft SDK requires WebRTC, my understanding is that it sends audio via standard websocket; and to get the microphone, my understanding is that it is doing something similar to your xrmp, meaning getUserMedia then Web Audio API:

[...]

PR #566, opened 2018-10-23
Fix jpeg capture and microphone on ML

machenmusik:
Incorporates the following patch commits:
- fix jpeg capture (918e60f and b15c824) - already in master
- update debug-ml.sh entrypoint (f19546d) - already in master
- there is no AudioContext reset, so don't call it (2af5ae2)

### DIFF
--- a/deps/exokit-bindings/webaudiocontext/src/AudioContext.cpp
+++ b/deps/exokit-bindings/webaudiocontext/src/AudioContext.cpp
@@ -15,6 +15,10 @@ lab::AudioContext *getDefaultAudioContext(float sampleRate) {
   return _defaultAudioContext.get();
 }
 
+void deleteDefaultAudioContext() {
+  _defaultAudioContext.reset();
+}
+
 AudioContext::AudioContext(float sampleRate) {
   audioContext = getDefaultAudioContext(sampleRate);
 }
@@ -31,7 +35,6 @@ Handle<Object> AudioContext::Initialize(Isolate *isolate, Local<Value> audioList
 
   // prototype
   Local<ObjectTemplate> proto = ctor->PrototypeTemplate();
-  Nan::SetMethod(proto, "close", Close);
   Nan::SetMethod(proto, "_decodeAudioDataSync", _DecodeAudioDataSync);
   Nan::SetMethod(proto, "createMediaElementSource", CreateMediaElementSource);
   Nan::SetMethod(proto, "createMediaStreamSource", CreateMediaStreamSource);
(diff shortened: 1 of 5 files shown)


machenmusik:
(Like the suspend-resume branch itself, this also requires the modulesio/LabSound#suspend-resume to be copied into native-audio-deps to compile.)

deps/exokit-bindings/webaudiocontext/src/AudioContext.cpp
```diff
@@ -9,14 +9,16 @@ lab::AudioContext *getDefaultAudioContext(float sampleRate) {
     _defaultAudioContext = lab::MakeRealtimeAudioContext(sampleRate);
 
     atexit([]() {
-      _defaultAudioContext.reset();
+      // NOTE: There was no reset method for AudioContext!
+      //_defaultAudioContext.reset();
```
modulesio: My thinking is we should reset the pointer to call the `AudioContext` destructor as intended. I think the issue is simply using the wrong signature, or wrong C++ version in compiler.

machenmusik:
So if I understand correctly, the intent is to delete and null the pointer, so reset needs to be called with nullptr as an argument, and that should work with whatever the current C++ version in use is?

modulesio:
Correct, we should leave it in and make sure it compiles. Was the problem that it wasn’t compiling?

machenmusik:
It did compile, but seemed to hang the thread that called it.

machenmusik:
The suggested change does not work; the original change (to not call reset) results in functional microphone behavior; the suggested change (to call reset with nullptr) still appears to hang the thread.

machenmusik:
I don't disagree, but that is outside the portions of the code with which I have sufficient familiarity at present.  Perhaps others can have a quick look at some point.

modulesio:
Yup, two other issues before I can loop back to this.

Maybe it makes the most sense to just continue the work from the `suspend-resume` branch?

machenmusik:
Yeah, this is a merge into `suspend-resume`, maybe you want to cherry-pick the three commits from master and not the commented-out reset -- again, sorry, I need to keep my branch intact, as it is functional enough for me at the moment

modulesio:
I rebased so `suspend-resume` contains the first three commits now.

I guess that means this PR can be closed?

machenmusik:
right, and you are planning to pursue a separate, more thorough fix for #477 and #537.

modulesio:
That wass the plan for the suspend-resume branch.

PR #567, opened 2018-10-24
Add onended for audio element

machenmusik:
Apologies, this is stacked upon #566 out of necessity; however this supplies an implementation of onended behavior for audio elements.  Perhaps you can cherry-pick the last two commits into a feature branch?


modulesio:
Moved to https://github.com/webmixedreality/exokit/pull/568.

machenmusik:
sorry about that, to test on my end I needed them stacked upon the other commits, but with the workaround #566 rejected, my branch will never merge cleanly with mainline.

PR #568, opened 2018-10-24
Add onended for audio element

modulesio:
Continuation of #567.


machenmusik:
LGTM, if it still works after the `suspend-resume` rebase...

modulesio:
- [ ] Should add a test for it in an audio example

modulesio:
The [newly added example](https://github.com/webmixedreality/exokit/pull/568/files#diff-be3eaa002fac063a7345497bbaf2f239R104) seems to fire correctly.

modulesio:
Tested both Windows and Magic Leap.

PR #571, opened 2018-10-27
Fix WebAudio SpeechRecognition polyfill

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/477.

Ingests https://github.com/modulesio/LabSound/pull/21.

The goal of this PR is to fix out the `WebAudio` features needed to bootstrap Speech Recognition on all Exokit platforms, including Magic Leap.

### DIFF
--- a/deps/exokit-bindings/webaudiocontext/include/AudioContext.h
+++ b/deps/exokit-bindings/webaudiocontext/include/AudioContext.h
@@ -74,7 +74,7 @@ protected:
   ~AudioContext();
 
 protected:
-  std::unique_ptr<AudioContext> audioContext;
+  std::unique_ptr<lab::AudioContext> audioContext;
 
   friend class Audio;
   friend class AudioListener;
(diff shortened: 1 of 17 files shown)


modulesio:
One remaining thing to do here is to ensure that multiple `AudioContext`s can be created. Currently we [only support one](https://github.com/webmixedreality/exokit/blob/eb3acf455bf9533b242f42a7357df8e939a028f1/deps/exokit-bindings/webaudiocontext/src/AudioContext.cpp#L7), for V1 laziness reasons.

However there is no good reason for this, and the SpeechRecognition polyfill takes liberties with creating new `AudioContext`s. This is probably a good place to fix the issue.

modulesio:
Note when debugging this: that polyfill has a message handling bug.

https://github.com/anteloe/speech-polyfill/issues/5

modulesio:
With the polyfill bug fix, the recognition seems to work well in my testing 👍.

PR #572, opened 2018-10-28
Bugfix LuminOS (ML) Canvas2D fonts

modulesio:
Magic Leap only supports one font at the moment (Lumino). It's available in the system image and is mounted, but Skia doesn't know how to pick it up by default.

This introduces a `fonts.xml` referencing the Lomino font, and a Skia update to load it.

Ingests https://github.com/modulesio/skia/pull/6.

Inspiration from [Servo's Lumin Runtime PR](https://github.com/servo/servo/pull/21985) for the `fonts.xml` trick, and @chrislatorres' [debug logging](https://github.com/modulesio/skia/pull/4/files).


modulesio:
Rebased + tested this in headset.

PR #575, opened 2018-10-28
Skia hardware accelleration

shaneharris:
Some stuff here might be useful, its only concerning GLFW so far and it doesn't work as of yet. After many hours of banging my head against the wall, i think it is beyond my capability currently. :(


modulesio:
Forked and continued in https://github.com/webmixedreality/exokit/pull/580.

Issue #80, opened 2018-05-30
Avatars conflict with each other


modulesio:
Having 3+ people in a XRMultiplayer can get into a state where players are manipulating each other's states, or a single player manipulates state for multiple players. Most likely some ID confusion.

modulesio:
Relevant repo (currently transcluded into exokit): http://github.com/webmixedreality/xrmp

modulesio:
A good code entry point would be https://github.com/webmixedreality/exokit-home/blob/master/index.html#L203

The 🐛 lies close to that code path.

modulesio:
Still reprod as of https://github.com/webmixedreality/exokit/commit/66237960f3060d3a1ad896f6839d51f9f45e7664, https://github.com/webmixedreality/exokit-home/commit/0d68fc76ad78cc2c351439c939421a8680e55829.

PR #577, opened 2018-10-29
2D Iframes with Servo2D/ChromiumEmbedded

modulesio:
This unlocks 2D iframes via CEF/Servo2D into Exokit..

capture6

The architecture here tries to mesh with Exokit's render loop by calling down to a hidden embedded browser instance for rendering, with Exokit driving the tick loop.

Unlike the electron experiment, this is all happening in the same process, different thread, with no transfers or IPC. Texture copy is minimal --

For CEF, we take the `OnPaint` callback and blast the subimage to texture, where the client can read it. It seems we can further optimize this to zero-copy by hijacking the DirectX external texture into OpenGL, at least on windows.

For Servo2D, we directly hijack the framebuffer and copy it to a texture when it's time to present a frame.

In the process this deletes the electron experiment.

Test API in `browser.html`:
```
  const gl = renderer.getContext();
  const tex = gl.createTexture();
  brwsr = new Browser(gl, tex, window.innerWidth, window.innerHeight, 'https://google.com');
```

The API could eventually just be an `iframe`, as we are doing with reality tabs. Then we can have both 2D and 3D iframes that work similarly in the XR space.

- [x] Add mouse/keyboard bindings
- [ ] Add browser destruction support
- [x] Clean up code
- [x] Figure out a cleaner API
- [ ] Add documentation
- [x] Add MacOS/Linux `native-browser-deps` CEF builds
- [x] Make it work on Magic Leap, if native web view support doesn't beat us to it


modulesio:
Branch for building CEF for arm64: https://github.com/modulesio/cef

#### Research

https://bitbucket.org/chromiumembedded/cef/issues/1990/linux-add-arm-build-support
https://bitbucket.org/chromiumembedded/cef/wiki/MasterBuildQuickStart

shaneharris:
ok after a bit of debugging i discovered if you use a webglrendertarget then the browser panel is black
uncomment the line below adding the object and it breaks
```<!doctype html>
<html>
<head>
    <meta charset="UTF-8">
    <script src="https://aframe.io/releases/0.8.2/aframe.js"></script>
    <script src="https://rawgit.com/feiss/aframe-environment-component/master/dist/aframe-environment-component.min.js"></script>
    <script src="https://threejs.org/examples/js/libs/inflate.min.js"></script>
    <script src="https://mattatz.github.io/THREE.Fire/assets/javascripts/FireShader.js"></script>
    <script src='https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.13.0/moment.min.js'></script>
    <script>
      AFRAME.registerComponent('test-browser', {
        init() {
          let iframe = document.createElement('iframe');
          iframe.d = 2;
          iframe.src = 'https://google.com';
          iframe.onload = () => {
              iframe.runJs(`document.body.style.background = '#000080';`);
          };
          const geometry = new THREE.PlaneBufferGeometry(1.2,1);
          const texture = new THREE.Texture();
          texture.repeat.set(1,-1);
          const properties = this.el.sceneEl.renderer.properties.get(texture);
          properties.__webglTexture = iframe.texture;
          properties.__webglInit = true;
          const material = new THREE.MeshBasicMaterial({
              map: texture,
          });
          const mesh = new THREE.Mesh(geometry, material);
          this.el.object3D.add(mesh);
          
          
      const geometry2 = new THREE.PlaneBufferGeometry(1.2,1);
        let renderTarget = new THREE.WebGLRenderTarget(512,512, { antialias: true } );
      const material2 = new THREE.MeshBasicMaterial({
              map: renderTarget.texture,
          });
      const mesh2 = new THREE.Mesh(geometry2, material2);
        //  this.el.object3D.add(mesh2);
        }
      });
    </script>
</head>
  <body>
    <a-scene background="color: black" keyboard-shortcuts="enterVR:false;exitVR:false" renderer="antialias: true; sortObjects: true; physicallyCorrectLights: true;"> 
      
      <a-entity test-browser position="0 1.6 -1"></a-entity>
    </a-scene>
  </body>
</html>```

modulesio:
Posterity: the above is actually a case of accidentally drawing black mesh on top of chromium embedded mesh. 😅

modulesio:
-  [ ] Need to add user-data-dir command line flag

This can reuse exokit’s data directory path, set table from the node command line.

modulesio:
Major magic leap android build tweaks for CEF `libcef.so` tracked here: https://github.com/modulesio/cef/pull/2

modulesio:
Re: https://github.com/webmixedreality/exokit/issues/604

I think the `iframe` webviews need a `profile: 'username'` key here, to prevent magical global state carrying over.

modulesio:
Known issue: mouse button back-click send causes a crash

PR #580, opened 2018-10-31
Skia hardware accelerated canvas

modulesio:
This PR adds OpenGL-backed `CanvasRenderingContext2D` support. The main reason is performance.

Since this transforms the canvas 2D context into an OpenGL texture, this unlocks CPU-less drawing and image manipulations and potentially running all media objects such as images/videos on the GPU. It also plays well with WebGL, since uploading textures with e.g. `gl.texImage2D` turns into a pure GPU texture copy. Finally, we elide some WebGL texture y-flipping during draw since this is all handled by the GPU code.

While this leverages the canvas2d texture for some things, it doesn't yet optimize things like `ImageData` or `Image`. However this PR is a prerequisite for that work.

The intent is for there to be no user change other than increased performance.

Forked from https://github.com/webmixedreality/exokit/pull/575.


shaneharris:
im seeing  `Failed to create CanvasRenderingContext2D surface` on `node . https://scrawny-buffalo.glitch.me/`

modulesio:
Yup, the case that's not covered is when the 2D canvas constructs before the 3D. It's a bug.

Thanks for the test case!

shaneharris:
Im seeing some images flipped on this branch, its intermittent and it seems its limited to JPG images
image
Notice the menu icons are PNG and i haven't seen an issue with those - the avatar images are all JPG and most of those come through as flipped - if i reload though it will change.

modulesio:
Thanks, do you know what the canvas 2d draw pattern is here?

shaneharris:
I think it could be related to the non POT textures, which i guess get resized though a canvas. Those will be POT in the end so that should sort itself out for me but useful to know. There is still some intermittentness to it aswell.

modulesio:
Ok great, then I'm thinking it's probably the resize itself that flips it.

shaneharris:
still flipped for me - 
image

shaneharris:
this is working really well for me now! i cant see anything else to do. Really nice work!

PR #581, opened 2018-11-01
Resource queueing for scripts and media

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/298.

This PR adds resource queuing to Exokit for media and scripts.

Previously Exokit would try to load all resources at the same time via node network calls. However, with some A-Frame examples it's possible for this to completely saturate the network (say, with 200 parallel image requests). The result in that case is timeouts or servers failing the requests.

The way that browsers handle this, and the thing this adds to Exokit, is resource queueing so that we only have a certain amount of resources in flight at a time. Only when a resource completes or fails do we continue to request the next one.

With this PR the parallel resource count has been capped to `8`, shared between scripts and media. This is an arbitrary number, but it seems to work ok.


modulesio:
Also added a massive script/image load test in `resources.html`.

modulesio:
Note that the [`maxParallelResources` const](https://github.com/webmixedreality/exokit/pull/581/files#diff-7eb52b366866677666470e019283c8eaR83) determines the degree of parallelism.

Although timeouts have still been seen with a 300-image load, this is a major improvement over the behavior. Therefore merging.

PR #583, opened 2018-11-03
Keyboard event updates - Shift Key Check, Capslock & e.code closer to browser

shaneharris:
Added shift key check in keyboard event to use the shift key value of keyboard layout instead of key value.

The e.key value was ';' on keyCode 186 for instance, compared to the browser it would be ':'


src/Event.js
```diff
@@ -107,7 +107,7 @@ class KeyboardEvent extends Event {
     if (init.key === undefined || init.code === undefined) {
       const keySpec = findKeySpecByKeyCode(init.keyCode);
       if (keySpec) {
-        init.key = keySpec.key;
+        init.key = init.shiftKey !== undefined?keySpec.shiftKey:keySpec.key;
```
modulesio: Can we clean up the spacing to be consistent with this file?

shaneharris: Yeh for sure, i just noticed this breaks with/without caps lock. i need to investigate more

shaneharris:
I could not find an easy way to get the initial caps lock state, so this just toggles it every time and event with keyCode 20 is fired. I think it should work OK like this, unless you think there is a better way?

modulesio:
No, I don't think this is the right path. I'm guessing there is a way to handle this in native GLFW, but I haven't done that research.

shaneharris:
Ok i added the caps + alpha check, I also changed it to take the keyCode straight from the keyboard layout file as that was again closer to the browser behavior and i also removed the first 10 entries from the keyboard layout for the digit keys - this allowed the shiftkey to behave correctly when pressing digit keys at the top but possibly means the same shift behavior for the numpad?? not sure i don't have a full keyboard to check handy.

modulesio:
I don't think any of this code should be in Javascript at all -- key handling is the role of the window system (GLFW in this case).

I suspect this doesn't detect an already-enabled capslock, or capslock triggered by the OS, amount other problems.

The correct fix is probably somewhere in the GLFW parameters or callbacks. Failing that, the fix should still probably be implemented close to GLFW bindings, not in the Javascript.

shaneharris:
Closing this as discussed 👍

modulesio:
Actually I think there was good code here, just in the wrong place...

PR #584, opened 2018-11-03
Cef keyinput modifiers

shaneharris:


package.json
```diff
@@ -81,7 +81,7 @@
     "window-fetch": "0.0.9",
     "window-selector": "0.0.5",
     "window-text-encoding": "0.0.2",
-    "window-worker": "0.0.97",
+    "window-worker": "0.0.98",
```
modulesio: Probably shouldn't include this in this PR. There is another PR where it's held up on multi-platform testing.

deps/exokit-bindings/browser/src/browser.cpp
```diff
@@ -536,19 +556,44 @@ NAN_METHOD(Browser::SendMouseWheel) {
   }
 }
 
+int GetKeyModifiers(Local<Object> modifiersObj){
+  int modifiers = 0;
+  if(modifiersObj->Get(JS_STR("shiftKey"))->BooleanValue()){
+    modifiers |= EVENTFLAG_SHIFT_DOWN;
+	std::cout << "shiftKey" << std::endl;
```
modulesio: Will have to remove these.

examples/browser.html
```diff
@@ -171,7 +171,7 @@
     const coord = planeMesh.projectMouse(e.clientX, e.clientY);
 
     if (coord) {
-      iframe.sendMouseDown(coord.x, coord.y);
+      iframe.sendMouseDown(coord.x, coord.y,0); //button 0 - left , 1 middle, 2 - right
```
modulesio: Should get the actual button from the event.

deps/exokit-bindings/browser/src/browser.cpp
```diff
@@ -578,18 +627,25 @@ NAN_METHOD(Browser::SendKeyUp) {
...
 NAN_METHOD(Browser::SendKeyPress) {
-  if (info[0]->IsNumber()) {
+  // Nan::HandleScope scope;
+  if (info[0]->IsNumber() && info[1]->IsObject()) {
     Browser *browser = ObjectWrap::Unwrap<Browser>(info.This());
     int key = info[0]->Uint32Value();
+	
+	Local<Object> modifiersObj = Local<Object>::Cast(info[1]);
+	int modifiers = GetKeyModifiers(modifiersObj);
+	if(modifiersObj->Get(JS_STR("shiftKey"))->BooleanValue()){
```
modulesio: This is the current bug: CEF not handling shift.

deps/exokit-bindings/browser/src/browser.cpp
shaneharris: will remove now

package.json
shaneharris: yep sorry about that one

deps/exokit-bindings/browser/src/browser.cpp
shaneharris: this does hit this block ok, so this part seems right

examples/browser.html
```diff
@@ -197,21 +197,21 @@
...
+      iframe.sendKeyUp(e.keyCode,{shiftKey:e.shiftKey,ctrlKey:e.ctrlKey,altKey:e.altKey});
     }
   });
   window.addEventListener('keypress', e => {
     if (focused) {
       if (e.keyCode === 114 && e.ctrlKey) { // Ctrl-R
         iframe.reload();
       } else {
-        iframe.sendKeyPress(e.keyCode);
+        iframe.sendKeyPress(e.keyCode,{shiftKey:false,ctrlKey:false,altKey:false});
```
modulesio: Wait why false?

PR #587, opened 2018-11-04
Key Input Modifiers & Mouse Buttons for CEF

shaneharris:
Also browser example updated.

### DIFF
--- a/deps/exokit-bindings/browser/src/browser.cpp
+++ b/deps/exokit-bindings/browser/src/browser.cpp
@@ -556,19 +556,36 @@ NAN_METHOD(Browser::SendMouseMove) {
   }
 }
 
+CefBrowserHost::MouseButtonType GetMouseButton(int button){
+	CefBrowserHost::MouseButtonType mouseButton;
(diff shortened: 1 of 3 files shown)


deps/exokit-bindings/browser/src/browser.cpp
```diff
@@ -615,19 +634,76 @@ NAN_METHOD(Browser::SendMouseWheel) {
...
+    modifiers |= EVENTFLAG_CONTROL_DOWN;// EVENTFLAG_COMMAND_DOWN  for mac?
+  }
+	
+  if(modifiersObj->Get(JS_STR("altKey"))->BooleanValue()){
+    modifiers |= EVENTFLAG_ALT_DOWN;
+  }
+  return modifiers;
+}
+
+map<int, int> keyCodesMap{
```
modulesio: Thanks for doing this 👍.

modulesio:
Actually... this is just going to the cef`/Chromium Embedded branch so I can test there 😅.

PR #588, opened 2018-11-05
added extra keypress event for keyCode 13 ( enter key ) it never gets fired otherwise.

shaneharris:
Essentially I forgot to test the enter key yesterday, I realized it didn't get fired because there was no `keypress` event for keyCode 13. This just sends the extra `keypress` event in that case and solves the issue.

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -2072,6 +2072,9 @@ class HTMLIFrameElement extends HTMLSrcableElement {
   }
   sendKeyDown(key, modifiers) {
     this.browser && this.browser.sendKeyDown(key, modifiers);
+	if(key === 13){
+		this.browser.sendKeyPress(key, modifiers);
+	}
   }
   sendKeyUp(key, modifiers) {
     this.browser && this.browser.sendKeyUp(key, modifiers);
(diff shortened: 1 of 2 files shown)


shaneharris:
I had considered adding this in the JS side with something like 
```javascript
sendKeyDown(key, modifiers) {
    this.browser && this.browser.sendKeyDown(key, modifiers);
	if(key === 13){
		this.browser.sendKeyPress(key, modifiers);
	}
  }
``` 
But i guess after our discussion on the previous PR its better to add this to the GLFW code instead. 

I haven't tested this on linux / mac either.

shaneharris:
cool beans will revert to the js solution!

src/DOM.js
```diff
@@ -2072,6 +2072,9 @@ class HTMLIFrameElement extends HTMLSrcableElement {
   }
   sendKeyDown(key, modifiers) {
     this.browser && this.browser.sendKeyDown(key, modifiers);
+	if(key === 13){
```
modulesio: It's helpful for these PRs to either set up the editor to match style of these files, or do a conversion after.

Otherwise things quickly become unreadable on Github.

shaneharris: yes i will do that ,

PR #590, opened 2018-11-06
iframe.contentWindow.location.href set in onloadend event

shaneharris:
Maybe this should have been in another PR but it only seemed to affect the browser so i said i would include it here - I added ensureCurrentGlWindow back into webgl.cc - maybe it was removed in a rebase regression? I'm probably making assumptions here and you removed it for good reason but i noticed the browser code depended on it and it was still in the header but not in the source. Anyway I couldn't see a good reason for its removal so i added it back in, hope i didn't over step!

Now for the main reason for this PR: 
Added in current URL to loadend event and dispatched on the front end on the iframe element as a `navigate` event - this way if the page loaded navigates somewhere else we get the URL of the new page passed back to the javascript side. Maybe I'm hijacking the loadend and this might be cleaner in its own event but i said i would show this working and not spend a lot of time on it if its not appropriate here.

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -1881,8 +1881,8 @@ class HTMLIFrameElement extends HTMLSrcableElement {
                   
                   let onmessage = null;
                   this.contentWindow = {
-                    location:{
-                      href:loadedUrl
+                    location: {
+                      href: loadedUrl
                     },
                     postMessage(m) {
                       browser.postMessage(JSON.stringify(m));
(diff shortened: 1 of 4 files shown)


shaneharris:
I also pushed a commit for a method `runCss` to inject css into the page. I guess this PR is the final requirements, but let me know if this should be broken up.

src/DOM.js
```diff
@@ -1877,6 +1881,9 @@ class HTMLIFrameElement extends HTMLSrcableElement {
                   
                   let onmessage = null;
                   this.contentWindow = {
+                    location:{
+                      href:loadedUrl
```
modulesio: Weird spacing

shaneharris: fixed

PR #591, opened 2018-11-06
Magic leap direct blit to headset

modulesio:
We use multisampled textures for all default framebuffers, and this requires a blit to a single-sampled texture before presenting to the headset APIs.

For presenting frames to the Magic Leap API, we were previously using an indirect texture blit to a local single-sampled texture, before blitting a second time to the device-provided textures. This does nothing except waste GPU time, and it's made slightly worse with Reality Tab blending since the draw calls are stacked up at the end of the frame and the final blit must wait for all draws to finish before adding its own overhead.

This PR shortcuts this intermediate blit for Magic Leap and composes the final framebuffer to the hardware-provided textures.


modulesio:
This works completely differently in the new reality tabs compositing strategy.

PR #592, opened 2018-11-07
Remove non-full builds from CI

modulesio:
Currently we have full and non-full builds being built on CI. Non-full contains just the results on `npm install`. Full contains a copy of `node` to run as well.

This PR removes the non-full version since they're not that useful -- you still need `node` and if you have node you might as well be building yourself.


modulesio:
These extra artifacts were [already removed in `master`](https://github.com/exokitxr/exokit/blob/764f44e3410d3afe15f4747478ebd2ff8f8bd322/.travis.yml).

PR #603, opened 2018-11-20
[WIP] Multithreaded rendering

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/382.
Fixes https://github.com/webmixedreality/exokit/issues/130.

This PR implements multithreaded rendering in Exokit. The goal is to break out every `window` context into its own thread of execution, which runs in parallel with all others.

## Reasoning

- **Bugs**: The first reason to do this is that the existing `window` and `Worker` implementations, [`vm-one`](https://github.com/modulesio/vm-one) and [`window-worker`](https://github.com/modulesio/window-worker), are custom hacks that reimplement the node tick loop. In addition to a history of platform-specific Worker bugs in implementing `importScripts`, this interacts badly with some native node code in e.g. the TLS implementation, necessitating more out-of-thread routing hacks. All of these problems go away if we just use first-class `worker_threads`, with native node support that does not need to be tracked or debugged.

- **Starvation**: The way the current tick loop works, the `window` tick execution is the only place that `window` contexts can make forward progress in `libuv`. That means that receiving request data is actually tied to the frame rate of the system -- this manifests as an inexplicable network slowness during loading. This is just how `uv_run` works by default, and trying to schedule work on the `window` instances is a problem that's 100% solved for us if we use `worker_threads`.

- **Performance**: The way the current tick loop works, we run each `window` serially in the main loop (no other way to do it, since all `window`s are bound to the same thread). By removing this restriction, we can run all `window` raf ticks, and therefore rendering, in parallel. This especially helps in the case of reality tabs, where we potentially have many render layers in the same renderloop, all of which can logically run independently. Even if you don't care about reality tabs, this is nonetheless a performance improvement for the single-window case, since Exokit's main loop itself can now run in parallel with your `window`.

- **Future features**: This paves the way for more futuristic features such as desynching misbehaving reality tab layers and correcting the blend with a userspace reprojection.

## Implementation

This requires [Node 11](https://github.com/webmixedreality/exokit/pull/601) for `worker_threads` stability. Most notably, this required a port of `libnode.a` for `arm64`/Magic Leap.

Additionally, this completely rewrites [`vm-one`](https://github.com/modulesio/vm-one) and [`window-worker`](https://github.com/modulesio/window-worker) in terms of `worker_threads`.

Finally, this PR requires some light synchronization where the Exokit renderloop needs to access/call into the rafs and events of the `window` instances.

### Dependent PRs

Ingests https://github.com/modulesio/window-worker/commit/ed59b8d9a30d7d2dec49a02097498bc6998faedb, https://github.com/modulesio/vm-one/pull/4.


modulesio:
- [ ] Still need to hook in the render loop's `window.requestAnimationFrame` map/join and global access

modulesio:
The work in this branch is outdated. All of this functionality has moved forward with https://github.com/exokitxr/exokit/pull/760.

PR #607, opened 2018-11-23
[WIP] Floor origin for AR

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/391.
Fixes https://github.com/webmixedreality/exokit/issues/506.

WebXR specifies that the origin of the XR coordinate frame should be the floor. This is not currently true (on Magic Leap), resulting in the sky content problem, which many people just hack around by moving content down.

This is suboptimal not only because it requires content to change based on AR user agent, but it also means that content written for VR sometimes needs tweaks for AR, which it shouldn't.

This PR uses the AR render loop to detect the floor and orient the XR origin to be at the largest "floor" type plane we see.

### DIFF
--- a/deps/exokit-bindings/magicleap/include/magicleap.h
+++ b/deps/exokit-bindings/magicleap/include/magicleap.h
@@ -116,7 +116,7 @@ class MLMesher : public ObjectWrap {
 public:
   static Local<Function> Initialize(Isolate *isolate);
 
-  MLMesher();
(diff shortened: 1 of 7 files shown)


modulesio:
This one has gotten far enough out of sync that it might make sense to start again.

Note that a lot of this work (possibly most of it) already made it to `master` since it relates to multi-tab XR offset transform application.

modulesio:
Closing in favor of https://github.com/webmixedreality/exokit/pull/698.

PR #610, opened 2018-11-24
WebXR hit test API

modulesio:
This updates the WebXR implementation to use [XRRay](https://immersive-web.github.io/webxr/#xrray-interface) where specced, which unlocks the `XRSession.requestHitTest()` API for use.

Implemented via meshing on Magic Leap.


modulesio:
Code should all be in now.

modulesio:
- [ ] Needs to be rebased

modulesio:
🎉

PR #612, opened 2018-11-24
Magic Leap MPK packaging

modulesio:
This PR adds initial support for building a custom MPK with user app content injected.


modulesio:
This needs to handle signing certificate injection as well.

PR #617, opened 2018-11-25
Remove electron integration

modulesio:
This removes the experimental `electron` integration in Exokit.

We have a faster and more robust solution with [Chromium Embedded Framework](https://github.com/webmixedreality/exokit/pull/577) for 2d `iframe`-to-texture. That approach is in-process (different thread) as opposed to this one, which relies on IPC, and goes directly to a GL texture, which `electron` never did.

If you were using this, please move your implementation over to the CEF `iframe` API.


modulesio:
Posterity: replaced with Chromium embedded framework.

PR #620, opened 2018-11-28
Fix OS X app bundle

shawwn:


modulesio:
Thanks, I think this regressed with https://github.com/webmixedreality/exokit/pull/459.

PR #621, opened 2018-11-28
Various ESLint fixes for src/index.js

shawwn:


modulesio:
This looks good, but there are major PR's floating right now that this might interfere with.

Let's hold off with merging this.

PR #625, opened 2018-12-04
Add --download support

modulesio:
This adds site download support to Exokit. With the `--download <directory>` option, Exokit will dynamically save all `GET`s it downloads via `fetch`/`XHR`. This isn't a static download like wget -- it actually executes the Javascript.

This is useful for packaging support https://github.com/webmixedreality/exokit/pull/612 so we can get all site resources for a URL.

This is basically another take on `--mirror` (https://github.com/webmixedreality/exokit/pull/622).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 node_modules
-build
+/build
 package-lock.json
 CMakeFiles
 CMakeCache.txt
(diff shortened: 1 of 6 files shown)


modulesio:
This works, other than requiring a `DISPLAY` head for the engine to run.

PR #629, opened 2018-12-07
Fix onhands object name

3DataWill:
On testing this appears to be `onhands` in the returned MLHandTracker object


modulesio:
[You are correct](https://github.com/webmixedreality/exokit/blob/bd39b1abd73bfb1c1cf156440e2892a271c8f6be/deps/exokit-bindings/magicleap/src/magicleap.cc#L600), thank you.

PR #646, opened 2018-12-16
Bump EGL pixel depth to RGBA8

modulesio:
Previously we were creating a 16-bit RGB context in EGL's `CreateNativeWindow`. This changes it into an RGBA8-backed context.

I'm not actually sure if this makes a difference to the final blit of the device (e.g. if it only handles RGB16), but this might be a good change for devices that _do_ use RGBA8 natively (e.g. Android ArCore).

This change continues to work on Magic Leap.


modulesio:
Too speculative for now. Let's keep the existing pixel configuration, since it's pulled from the official SDK examples.

PR #649, opened 2018-12-17
Reality tabs minimap example

modulesio:
This adds emulated support for the IRL minimap example as a Reality Tab in the realitytabs.html example.


modulesio:
Currently working in fake display mode on desktop. Magic leap reality tabs layers are affected by an unrelated render target set framebuffer dangling problem.

Merging the checkpoint -- this is still not the default home load.

PR #658, opened 2018-12-19
XR offset support in hardware APIs

modulesio:
For reality tabs, as well as the [WebXR spec](https://github.com/immersive-web/webxr/pull/441), we need to implement XR origin transformations via `XRRigidTransform`.

This is already done in the reality tabs blend/offset case, but we need to do it for all hardware apis exposed, including hit testing, meshing, hands tracking, image tracking, object detection and others.

- Normalize transforms on the `XRRigidTransform` API
- Add `XRRigidTransform` output to hardware binding APIs (particularly magic leap)
- Update examples to use the `XRRigidTransform`
- Update API docs to reflect the transforms output


modulesio:
- [ ] From testing this has a bug where planes normals are always axis-aligned, which is not correct

PR #659, opened 2018-12-19
Per-OS native-browser-deps

modulesio:
For Exokit's iframe-to-texture support (backed by CEF/Servo2D), we need to include some massive (relatively speaking) browser dependency libraries, for all operating systems.

The problem is that these are too big to fit on NPM, and even Github won't accept some of them (Linux) unless manually split.

This breaks the dependencies into per-OS npm modules that _should_ fit.


modulesio:
This is related to https://github.com/webmixedreality/exokit/pull/625, since installing exokit as a pure scripting/download solution requires this -- `npm` bugs out and refuses to install exokit as a dependency otherwise.

PR #672, opened 2018-12-27
Add monospace fonts

modulesio:
On non-desktop systems (Magic Leap/Android), fonts availability is controlled via `fonts.xml`. Right now we just use the single system font, but we can theoretically support any font we can get a file for.

One notable missing font is a `monospace` one, which is required for in-XR devtools rendering.

This PR adds [Inconsolata](https://fonts.google.com/specimen/Inconsolata) as the default monospace font for non-desktop Exokit.


modulesio:
Tested and works in `<canvas>`.

There is an unrelated bug in `CanvasRenderingContext2D.measureText` -- it seems to use metrics of the old font that was set.

PR #677, opened 2018-12-29
Fix crash in image-context

j-devel:
I have observed frequent crashes of Exokit when the web app is dealing with ajax-loaded images.  The following example reproduces the crash fairly reliably (not always though; maybe try several times in case the crash does not manifest in one shot).

```
$ node . -h https://j-devel.github.io/pixels-to-mesh-test/index.html
```

**Cause of the problem**

In `image-context.cc`, the `handle` in `uv_close()` points to `threadAsync` of `Image`.  But this address `&threadAsync` is no longer valid after the `Image` instance is destroyed.  Depending on the situations, `Image` destruction occurs _before_ the actual `uv_close()` operation is fulfilled, leading to unpredictable behaviors (mostly crashes with an assertion failure of `uv__finish_close`), e.g.

```
Assertion failed: (handle->flags & UV_HANDLE_CLOSING), function uv__finish_close, file ../deps/uv/src/unix/core.c, line 252.
Abort trap: 6
```

**Fix**

This PR is to allocate the async handle in heap and let `uv_close()` work properly without causing the crash.

### DIFF
--- a/deps/exokit-bindings/canvascontext/src/image-context.cc
+++ b/deps/exokit-bindings/canvascontext/src/image-context.cc
@@ -68,7 +68,9 @@ void Image::RunInMainThread(uv_async_t *handle) {
   image->arrayBuffer.Reset();
   image->error = "";
 
-  uv_close((uv_handle_t *)handle, nullptr);
+  uv_close((uv_handle_t *)handle, [](uv_handle_t *handle) {
+    free(handle);
+  });
 }
 
 void Image::Load(Local<ArrayBuffer> arrayBuffer, size_t byteOffset, size_t byteLength, Local<Function> cbFn) {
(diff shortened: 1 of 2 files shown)


modulesio:
Thanks, great catch!

deps/exokit-bindings/canvascontext/src/image-context.cc
```diff
@@ -79,9 +81,10 @@ void Image::Load(Local<ArrayBuffer> arrayBuffer, size_t byteOffset, size_t byteL
     this->cbFn.Reset(cbFn);
     this->error = "";
 
-    uv_async_init(uv_default_loop(), &threadAsync, RunInMainThread);
+    threadAsyncHandle = (uv_async_t *)malloc(sizeof(uv_async_t));
```
modulesio: I'm partial to `new`/`delete`, since this is all C++ code. In the future it could be a smart pointer.

PR #682, opened 2018-12-30
added loop get/set for HTMLAudioElement

shaneharris:
tested and working.

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -2583,6 +2583,16 @@ class HTMLAudioElement extends HTMLMediaElement {
     }
   }
 
+  get loop() {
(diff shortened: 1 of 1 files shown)


src/DOM.js
```diff
@@ -2583,6 +2583,16 @@ class HTMLAudioElement extends HTMLMediaElement {
     }
   }
 
+  get loop() {
+    return this.audio && this.audio.loop;
```
modulesio: Could it be:

```
return this.audio ? this.audio.loop : false;
```

I don't think `undefined` is a legal value if we ever end up seeing it.

modulesio: I'll merge + patch this.

PR #697, opened 2019-01-04
Async ( sync ) clipboard support

shaneharris:
`https://developer.mozilla.org/en-US/docs/Web/API/Clipboard`

This PR adds partial support for the async ( still sync under the hood ) clipboard API with the clipboard property exposed on the navigator object.

```javascript  
navigator.clipboard.readText()
  .then(text => {
    // console.log(text);
  })
```

```javascript  
navigator.clipboard.writeText(value)
  .then(()=> {
    // Text is in clipboard
  })
```

I have removed the previous (slightly modified ) code blocks responsible for copy / paste --

```javascript
window.document.on('paste', e => {
  e.clipboardData = new window.DataTransfer();
  contexts.forEach(context => {
    const windowHandle = context.getWindowHandle();
    const clipboardContents = nativeWindow.getClipboard(windowHandle).slice(0, 256);
    const dataTransferItem = new window.DataTransferItem('string', 'text/plain', clipboardContents);
    e.clipboardData.items.push(dataTransferItem);
  });
});
```


shaneharris:
https://github.com/webmixedreality/exokit/blob/3348908d00b08f63c26f0502992d6e5f87eeba7f/src/Document.js#L102-L108
Also not sure if this is needed if the PR looks good.

src/index.js
```diff
@@ -909,17 +909,6 @@ const _bindWindow = (window, newWindowCb) => {
   window.on('unload', () => {
     clearTimeout(timeout);
   });
-  window.on('navigate', newWindowCb);
```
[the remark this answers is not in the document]
shaneharris: should i put the `window.document.on('paste',....` block back too?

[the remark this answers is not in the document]
modulesio: I think the navigate event is required to act on window navigation.

Also, the `document` `paste` should have already been working, but that only made sense in the context of https://github.com/webmixedreality/exokit/pull/697/files#diff-614df7e2218ad0ff0e6a77be196d9f7cR107.

So I understand the clipboard API addition (thanks!), but I don't get the removals.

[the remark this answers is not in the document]
shaneharris: you know what i must have been in a weird mood when i did that. I have undone all the removals. sorry about that.

deps/exokit-bindings/glfw/src/glfw.cc
```diff
@@ -1351,8 +1351,16 @@ NAN_METHOD(GetClipboard) {
 
 NAN_METHOD(SetClipboard) {
   NATIVEwindow *window = (NATIVEwindow *)arrayToPointer(Local<Array>::Cast(info[0]));
-  String::Utf8Value str(info[0]->ToString());
-  glfwSetClipboardString(window, *str);
+  if (info.Length() >= 2) {
+    // Don't want to risk passing a null string around, which might create unpredictable behavior.
```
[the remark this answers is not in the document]
shaneharris: Good question! That code actually came straight from a previous PR. I just copy pastad that for the most part

src/core.js
```diff
@@ -983,6 +983,24 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
       return fakeVrDisplay;
     },
     getGamepads,
+    clipboard:{
+      read:() => Promise.resolve(), // Not implemented yet
+      readText: () => new Promise(resolve => {
+        if (GlobalContext.contexts && GlobalContext.contexts.length){
```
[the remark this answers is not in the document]
shaneharris: Yep let me add an else & reject there

shaneharris:
any more changes needed here?

src/core.js
```diff
@@ -983,6 +983,28 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
...
     getGamepads,
+    clipboard:{
+      read:() => Promise.resolve(), // Not implemented yet
+      readText: () => new Promise((resolve, reject) => {
+        if (GlobalContext.contexts && GlobalContext.contexts.length){
+          const windowHandle = GlobalContext.contexts[0].getWindowHandle();
+          const clipboardContents = nativeWindow.getClipboard(windowHandle).slice(0, 256);// why do we slice this?
+          resolve(clipboardContents);
+        }else{
+          reject();
```
modulesio: This would be a puzzling error to recieve.

deps/exokit-bindings/glfw/src/glfw.cc
```diff
@@ -1351,8 +1351,15 @@ NAN_METHOD(GetClipboard) {
 
 NAN_METHOD(SetClipboard) {
   NATIVEwindow *window = (NATIVEwindow *)arrayToPointer(Local<Array>::Cast(info[0]));
-  String::Utf8Value str(info[0]->ToString());
-  glfwSetClipboardString(window, *str);
+  if (info.Length() >= 2) {
+    Nan::Utf8String utf8_value(info[1]);
+    int len = utf8_value.length();
+    std::string name(*utf8_value, len);
```
modulesio: What is this line for?

src/core.js
shaneharris: change to `reject(new Error('No context available.'));` instead?

deps/exokit-bindings/glfw/src/glfw.cc
shaneharris: Im not sure TBH, it worked so i didn't change it from what it was:

would this work better:

```
NAN_METHOD(SetClipboard) {
  if (info.Length() >= 2) {
    NATIVEwindow *window = (NATIVEwindow *)arrayToPointer(Local<Array>::Cast(info[0]));
    String::Utf8Value str(info[1]->ToString());
    glfwSetClipboardString(window, str);   
  } else {
      return Nan::ThrowTypeError(
            "must provide arg length > 1");
  }
}
```

src/core.js
modulesio: That would be ok, but this brings up the question of why a gl context is needed to read clipboard. Is there any way to do it without a window/canvas?

shaneharris: Indeed! I have replaced the passing of the window context in favor of `GetCurrentWindowContext` which removes the need for the error condition at all.

deps/exokit-bindings/glfw/src/glfw.cc
shaneharris: oh yes duh, i see what you mean now. I'm a doofus. Ok how does the latest look?

shaneharris:
Need anything more from me on this one?

modulesio:
Don’t worry, this isn’t even top 100 most painful reviews.

deps/exokit-bindings/glfw/src/glfw.cc
```diff
@@ -1349,10 +1349,15 @@ NAN_METHOD(GetClipboard) {
...
 NAN_METHOD(SetClipboard) {
-  NATIVEwindow *window = (NATIVEwindow *)arrayToPointer(Local<Array>::Cast(info[0]));
-  String::Utf8Value str(info[0]->ToString());
-  glfwSetClipboardString(window, *str);
+  NATIVEwindow *window = GetCurrentWindowContext();
+  if (info.Length() == 1) {
+    Nan::Utf8String utf8_value(info[0]);
+    glfwSetClipboardString(window, *utf8_value);   
+  } else {
+      return Nan::ThrowTypeError("must provide arg length > 0");
```
modulesio: This error is not in line with the if. Can we just say “invalid arguments”? The stack will point here for anyone who cares to debug.

shaneharris: Done!

PR #705, opened 2019-01-07
<canvas> toDataURL/toBlob base64

modulesio:
Fixes https://github.com/webmixedreality/exokit/issues/704.

Previously we were writing a binary string in the `toDataURL` codepath. This PR makes Skia return an `ArrayBuffer` instead, which we consume with `toDataURL` and convert to `base64` there.

Additionally, this adds `toBlob` support since with this architecture it was straightforward to implement.


shaneharris:
Tested this and both the toDataURL and toBlob are working well!

modulesio:
Sweet, first try. Thanks for testing!

PR #711, opened 2019-01-07
Update BUILDING.md Magic Leap signed instructions

chrislatorres:
The build-ml.sh `--signed` arg was previously removed ( https://github.com/webmixedreality/exokit/commit/b6c17b24387eae47ecd76a8c11204741700b1f4f#diff-9a4c9d3e3677de278c34f29937c44473L61 ). 

The build-ml script now signs by default ( https://github.com/webmixedreality/exokit/blob/master/scripts/build-ml.sh#L56 ).


modulesio:
Thanks!

PR #714, opened 2019-01-08
Basic reality tab scrolling and scaling

chrislatorres:
Adds basic scrolling on the z axis to reality tabs using the touchpad's y axis.


modulesio:
This seems like a good start but I think we can do better without much effort.

From my reading of this code it’s taking the absolute pad value which limits the range. But if we accumulate a drag, we could get unlimited swipe range.

Check out the tabs/links scrolling code for the reality tab listMesh. That’s probably stealable...

chrislatorres:
e4f8175 adds drag accumulation. It could still be improved in that it will continue moving on the z axis  when the pad is unpressed.

chrislatorres:
>  can we make the scroll tracked per-controller in an array? Even just for magic leap, having two controllers at a time would make this behave jankily otherwise.
  
c6a7234
  
> I think the scroll can go negative here. Can we Math.max(x, 0)?
  
cc1a9e2

modulesio:
For the first change it looks like there is still only a single scroll
accumulator, for both controllers.

Also, a better way to cap to zero would be in the accumulator. Otherwise
you can scroll into the negatives and it will appear stuck at zero until
you I scroll all the way back.

chrislatorres:
When scrolling an iframe, the "Tabs" menu will also scroll up/down from the wheel event listener: 

https://github.com/webmixedreality/exokit/blob/c6a7234104782054b26277b36757a30c310d6689/examples/realitytabs.html#L1219-L1223

chrislatorres:
> For the first change it looks like there is still only a single scroll accumulator, for both controllers.

Should there be multiple scroll accumulators? If there are 2 controllers, and `scrollAxis[0] = 12` and `scrollAxis[1] = 0`, the iframe will stay at 0.

```
      for (let i = 0; i < gamepads.length; i++){
        const gamepad = gamepads[i];
        if(scrollAxis[i] ==  null){
          console.log('scrollAxis is null');
          scrollAxis[i] = gamepad.axes[1] * 20;
        }
        scrollAxis[i] += gamepad.axes[1] / 20;
        console.log('scrollAxis = ' + scrollAxis[i]);
        //scrollAxis[i] += Math.max(gamepad.axes[1] / 20, -1.51);

        localMatrix
          .fromArray(pose.targetRay.transformMatrix)
          .decompose(localVector, localQuaternion, localVector2);

        localVector
          .add(
            localVector2.set(0, 0, -1.5 - scrollAxis[i])
            .applyQuaternion(localQuaternion)
          );
        iframe.position = localVector.toArray();
        iframe.orientation = localQuaternion.toArray();
      }
```

Log example:
screenshot from 2019-01-09 13-35-45

chrislatorres:
> Also, a better way to cap to zero would be in the accumulator. Otherwise you can scroll into the negatives and it will appear stuck at zero until you I scroll all the way back.
  
f8c463e

> Perhaps listen for the pad higher up and depending on whether we are in move mode for that controller, handle the change accordingly?
  
1e62066

chrislatorres:
82232d7 adds scaling functionality on the gamepad's X axis (left/right) (down-scale/up-scale), this is a best practice for XR.

examples/realitytabs.html
```diff
@@ -1767,6 +1775,7 @@
                         const pressed = gamepad.buttons[1].pressed;
                         const lastPressed = lastPresseds[i];
                         if (pressed && !lastPressed) {
+                          moveMode = true;
```
modulesio: Move mode is already tracked in `moves[i]`, per-controller, so why nod just use that [here](https://github.com/webmixedreality/exokit/pull/714/files#diff-bdc418c11110bb6bec917c75b88495f9R1898)?

examples/realitytabs.html
```diff
@@ -1398,19 +1398,26 @@
...
+	  .decompose(localVector, localQuaternion, localVector2);
+
+        localVector
+   	  .add(
+    	    localVector2.set(0, 0, -1.5 - scrollAxis)
+  	    .applyQuaternion(localQuaternion)
+	  );
+	iframe.position = localVector.toArray();
+	iframe.orientation = localQuaternion.toArray();
+        iframe.scale = scale.toArray();
```
modulesio: Weird spacing.

examples/realitytabs.html
chrislatorres: fixed 
f3848f1

examples/realitytabs.html
chrislatorres: fixed 
f3848f1

chrislatorres:
https://github.com/webmixedreality/exokit/pull/714/files#diff-bdc418c11110bb6bec917c75b88495f9R1406

Maybe it should be possible to scale below 1, as in: .25 or .5 scaling. 2d iframes still appear large at 1x size.

modulesio:
Yes it should be possible to go down to very small.

Though if the 2d iframes are too large, we can fix that by changing the default geometry size. I actually agree it's probably too big/too close right now.

chrislatorres:
`scrollAxis` and `scaleAxis` should be tracked per reality tab, otherwise the scroll/scale will carry over when moving a different tab.

Looking for a way to uniquely identify each tab, and then use the identifier as a key for `scrollAxis` and `scaleAxis` as objects, with the key being RT identifier and value being axis.

for example:
`scrollAxis[tab] = Math.max(scrollAxis[tab] + gamepad.axes[1] / 20, 0);`
`scaleAxis[tab] = Math.max(scaleAxis[tab] + gamepad.axes[0] / 20, .15);`

modulesio:
Why did reality `projection_ml.html` make it in here?

modulesio:
I merged `master` into here and deleted `projection_ml.html` because I don't think it had anything to do with this PR.

Hopefully that's ok; it's in the history if it's needed for something.

chrislatorres:
> Why did reality realityprojection_ml.html make it in here?
  
That was a mistake, must have done a `git add realityprojection_ml.html` by accident.

modulesio:
Continued in https://github.com/webmixedreality/exokit/pull/734.

PR #717, opened 2019-01-09
Reality tabs tutorial

modulesio:
This PR makes it so that instead of loading the exobot HTML page with no explanation, the default reality tab becomes a tutorial for how to open and experience the Exokit reality tabs examples.

Additionally, auxiliary rendering such as the initial floor are moved to the tutorial reality tab, where it can be deleted.


chrislatorres:
How should the tutorial be presented? Text-based, spatial audio, exobot-[personality](https://youtu.be/9jpUVVJ7CVQ?t=14)?

Thoughts:
- Prompt upon first open of Exokit (Would you like to view tutorial? Y/n) and remember for subsequent opens of Exokit

- Exobot guiding through a sequence of animations per topic. Each topic/sequence as a discrete section to swipe left/right (backward/forward) through topics.

- Is there a way to detect, as part of the tutorial, that a user has clicked into the correct section in the realitytabs menu?

PR #722, opened 2019-01-11
Reality tabs multiplayer

modulesio:
This PR adds a thin multiplayer layer to the reality tabs example. The result allows placing and traversing 3D+2D WebXR sites in a multiplayer party. The implementation is itself just WebXR so it's device agnostic.

The backing server protocol is the dumb [XRMP client](https://github.com/webmixedreality/xrmp) and [server](https://github.com/webmixedreality/multiplayer).

The point of this is to demonstrate how a full "browser" experience re-envisioned with reality tabs could look like, powered by Exokit engine. Once this stabilizes we'll probably break out the reality tabs browser to a different repo, and keep the engine (this repo) separate.


modulesio:
Still needs more testing but this should be allowing casting of shared WebXR reality tabs for a simple party now.

capture 11

PR #726, opened 2019-01-11
Bugfix unconnected gamepads flagged as connected

### LINKED ISSUE #723: Reporting non-existing controllers
Currently `gamepads.length` = 2 and `inputSources.length` = 2 because 2 controllers are reported by default:

https://github.com/webmixedreality/exokit/blob/1956ca422ed86842370163abb27311948c101119/src/core.js#L1770-L1771
-->
https://github.com/webmixedreality/exokit/blob/76e5d7888dcf49487288cc185c88e4503b45fef3/src/VR.js#L487-L499  
  
and
  
https://github.com/webmixedreality/exokit/blob/726a0d5e3247c80c048c65875a0eb49392efc1d7/src/XR.js#L118-L121
  

Visual of extra controllerMesh/rayMesh in an example due to non-existing controller reported:
dwmwd-qvaaertem

modulesio:
Fixes #723.

Same thing as https://github.com/webmixedreality/exokit/pull/725.


modulesio:
Updated the examples. This needs to be tested.

PR #737, opened 2019-01-14
Reality tabs depth population example

modulesio:
Experimental attempt at moving depth population from the reality tabs environment by default, to an example reality tab we can open.

A better way to do this might be to expose an API for each reality tab to individually decide whether it wants its depth buffer populated or not.


chrislatorres:
This is a better design, a related issue I had: when opening a `camera_ml.html` reality tab to stream the MLO camera, the output for the `camera_client.html` with depth population **enabled** was black. 
  
This is probably due to the nature of how additive displays use black for occlusion.

-------------------------------------
  
Visual of depth-population enabled vs disabled:
screenshot from 2019-01-09 16-38-38
screenshot from 2019-01-09 16-54-13

modulesio:
No I think that would just be an Exokit bug, since depth population is done via a render of black (I think), There are better ways to do it, including an actual API on ML platform (file:///C:/Users/avaer/MagicLeap/mlsdk/v0.19.0/docs/api/html/ml__graphics__utils_8h.html).

PR #748, opened 2019-01-19
Multithread workers

modulesio:
This ports exokit's web worker implementation directly to `worker_thread` now that we have node 11. Fixes a crash with the previous threading hack, because that implementation is deleted.

This is the first of two multithreading ports; the second is to do the same thing with the Window implementation.

Fixes https://github.com/webmixedreality/exokit/issues/746.


modulesio:
Note also that this now means that running with `--experimental-worker` is required. However, we attempt to enforce this on startup.

modulesio:
This fixed The Expanse crashing for me.

PR #754, opened 2019-01-21
Update OpenVR

modulesio:
This updates Exokit's OpenVR bindings to `1.1.3b`. Steam recently updated their code and Oculus shim support broke and being up to date is probably the first step of fixing that.

Attempted fix for https://github.com/webmixedreality/exokit/issues/742.


modulesio:
Works on OpenVR.

PR #757, opened 2019-01-21
Modular native bindings

modulesio:
In preparation of parallel user `window`s, we need to encapsulate native bindings in a module that can be required from `worker_threads`.

Since we currently do a `setNativeBindingsModule` hack to latch the native bindings at the top context, this needs to be refactored into a self-contained bindings module/file. That's what this PR does.


modulesio:
We should use the `Worker` implementation as a test bed for the bindings working across `worker_threads` -- since `Worker` is already `worker_threads`-based.

PR #759, opened 2019-01-23
Pre-parallel renderloop

modulesio:
Exokit's render loop comes from an era where we didn't have reality tabs, and we weren't thinking about composing layers or threading. So, a lot of the XR code was built around the notion of ticking windows and contexts.

This made it hard to refactor and reason about the renderloop holistically, or do a map-reduce render flow like we need for parallel reality tabs rendering/composition.

This PR refactors the render loop into a single recursive method that breaks out ticks and blitting to the windows and contexts in a controller fashion, in preparation for running all user windows in parallel.

Additionally, this refactors tracking XR state to a single abstract xr state structure, which uses `SharedArrayBuffer` to communicate the XR viewer states to the XR device emulators.

This PR should additionally reduce memory usage (due to less entangled XR objects being created) and make for shorter/more optimizable V8 straight line code.

### DIFF
--- a/deps/exokit-bindings/magicleap/include/magicleap.h
+++ b/deps/exokit-bindings/magicleap/include/magicleap.h
@@ -294,6 +294,8 @@ public:
   static NAN_METHOD(SubmitFrame);
   static NAN_METHOD(IsPresent);
   static NAN_METHOD(IsSimulated);
+  static NAN_METHOD(GetSize);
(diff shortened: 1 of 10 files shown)


modulesio:
Tested on magic leap.

modulesio:
Still should test in VR.

modulesio:
Also tested on OpenVR. Small perf improvement from the single centralized renderloop 🎉.

PR #764, opened 2019-01-27
[ML] User control for GL buffers

modulesio:
The magic leap bindings previously dabbled in setting the GL buffers for features such as meshing, for performance reasons.

However, with user window parallelization, whose main point is to reduce GL context switching outside of user code, this makes a lot less sense. Additionally, the previous API led to user confusion re: buffer validity, and for at least one user this was a performance complaint.

Therefore, this PR drops this feature in favor of letting the user set the GL buffers themselves, if they so wish.

### DIFF
--- a/deps/exokit-bindings/magicleap/include/magicleap.h
+++ b/deps/exokit-bindings/magicleap/include/magicleap.h
@@ -132,14 +132,9 @@ public:
 
 class MeshBuffer {
 public:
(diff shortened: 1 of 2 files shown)


modulesio:
With this, documentation needs to be updated as well.

modulesio:
Closing in favor of https://github.com/exokitxr/exokit/pull/939/files.

PR #775, opened 2019-02-15
Add Query for WebGL 2

Palmer-JC:
Have not tested.  Am going to look into where the actual tests come from (repo) for [https://www.khronos.org/registry/webgl/sdk/tests/webgl-conformance-tests.html?version=2.0.1](https://www.khronos.org/registry/webgl/sdk/tests/webgl-conformance-tests.html?version=2.0.1)


Palmer-JC:
I do not really understand this checks stuff.  The log file is so big it will not even show in the browser.

I have found the place for the conformance tests specific to Query, [https://github.com/KhronosGroup/WebGL/tree/master/conformance-suites/2.0.0/conformance2/query](https://github.com/KhronosGroup/WebGL/tree/master/conformance-suites/2.0.0/conformance2/query).  Hopefully they will write their result to console.

deps/exokit-bindings/gluegen/glbindinit.h
```diff
@@ -399,3 +399,10 @@
 	_gl->Set(v8::String::NewFromUtf8(isolate, "vertexAttrib4f"), v8::FunctionTemplate::New(isolate, __vertexAttrib4f));
 	_gl->Set(v8::String::NewFromUtf8(isolate, "vertexAttribPointer"), v8::FunctionTemplate::New(isolate, __vertexAttribPointer));
 	_gl->Set(v8::String::NewFromUtf8(isolate, "viewport"), v8::FunctionTemplate::New(isolate, __viewport));
+	_gl->Set(v8::String::NewFromUtf8(isolate, "createQuery"), v8::FunctionTemplate::New(isolate, __createQuery));
```
modulesio: I don't think we need this, since this file is not currently used in the build (used to generate mobile bindings, which is a separate issue).

modulesio:
The check looks like a Linux run failure after building. Trying it on my VM.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4847,6 +4858,77 @@ NAN_METHOD(WebGL2RenderingContext::New) {
...
+
+  info.GetReturnValue().Set(queryObject);
+}
+
+NAN_METHOD(WebGL2RenderingContext::GetQueryParameter) { // adapted from GetProgramParameter
+  GLint queryId = info[0]->ToObject()->Get(JS_STR("id"))->Int32Value();
+  GLenum pname = info[1]->Int32Value();
+  
+  switch (pname) {
+    case GL_QUERY_RESULT_AVAILABLE:
```
modulesio: I think these constants also need to be exposed on the context: https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/getQueryParameter#Parameters

modulesio: The constants are all exposed here: https://github.com/webmixedreality/exokit/blob/6c9c7915cdf85bd99da8b16be775af06ebddf23c/deps/exokit-bindings/webglcontext/src/webgl.cc#L73

modulesio: We should make sure to get them all for the other functions as well, such as `beginQuery`: https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/beginQuery#Parameters

modulesio:
I managed to run this PR on Linux, so I think it might be an issue on Circle CI's end on Linux.

Palmer-JC:
Ok, I abandoned trying to use the conformance testing to do validation.  We only need to validate that the wrappers work, not that the context is correct.  Not our problem.  Those tests are not a good fit.

[http://webglsamples.org/WebGL2Samples/](http://webglsamples.org/WebGL2Samples/), [GitHub](https://github.com/WebGLSamples/WebGL2Samples), seem to have a test for all webgl2 stuff, and would seem a good basis for testing all new 2.0 functionality.

When trying the [query test](http://webglsamples.org/WebGL2Samples/#query_occlusion), it fails as you predicted because the gl constants have not been added.  I will now just grab all of them from the [WebGL2 standard](https://www.khronos.org/registry/webgl/specs/latest/2.0/).  Section 3.7 lists them all out.  When I do, and have the test for query working, I will commit to my fork on github, expect on Monday.  I have not ever figured out how to add another commit to a PR before though.

Also, I will throw in a few extra calls to test all of the wrapper-ed function, like isQuery that the sample did not have.

modulesio:
Yeah, conformance testing is a can of worms we don't have to open at this point.

Thanks for being thorough with the rest of the query functions!

> I have not ever figured out how to add another commit to a PR before though.

Pushing should be enough for it to appear here. 🎉

Palmer-JC:
Well I added everything from standard as constants, except those already in use.  I left the lines, just commented out.

It will not compile.  Thing sooner or later these need to be defined.  There are versions of some with an `_NV` at the end, but that is only as an extension.  WebGL2 got rid of a lot of extensions.

Tracked definitions down to being in [https://github.com/mikeseven/node-native-graphics-deps](https://github.com/mikeseven/node-native-graphics-deps).  Problem is this has not been updates in years, before WebGL2.  Please advise.

Palmer-JC:
Ah, I see you have a fork.  FYI, the entire header is in section 3.7 of standard linked 4 comment up.

Palmer-JC:
For now, I have stuffed all the definitions into webgl.h.  I get quite a few redefinition warnings, but it is building.  You can find the place you really want this stuff to be at your leisure.

The Query test page does not error, but it also does not draw a triangle.  Going to edit page for missing calls.

modulesio:
Probably the right place for all of the constants is indeed `webgl.h`.

However I'm guessing most of the C++ constants alread come in from the platform headers, so would not need to be manually defined. They only need to be added to the JS exposure on the user-facing object and there are `#define`s. to do that.

Ideally we would process that spec and come up with exposure code for all of those, but we don't necessarily need to declare any numbers ourselves.

Mind pushing what you have to `Palmer-JC:master`?

Palmer-JC:
Done

deps/exokit-bindings/webglcontext/include/webgl.h
```diff
@@ -85,6 +85,277 @@ enum GlKey {
   GL_KEY_PLANE,
 };
 
+#define GL_READ_BUFFER                                  0x0C02
```
modulesio: How many of these are actually needed? I suspect that the number is low since the platform is providing these. In that case, I would defer to the platform versions since those are the exact values we pass back to the platform when we do make WebGL calls.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -625,6 +626,279 @@ void setGlConstants(T &proto) {
 
   // external
   JS_GL_SET_CONSTANT("TEXTURE_EXTERNAL_OES", 0x8D65);
+
```
modulesio: Maybe it makes sense to delete the WebGL constants and just use this list for all of them?

Palmer-JC:
Your review revolves around where things are, and you can get this stuff where ever you wish.

I have been more concerned right now that a [very very simple page](https://github.com/WebGLSamples/WebGL2Samples/blob/master/samples/query_occlusion.html) does not work, even with all the query stuff commented out.  I think it has to do with the `OES_vertex_array_object` extension.  In WebGL2 this is not an extension, and needs to call ES 3.0 `glGenVertexArrays`.

Think converting any extensions which are no longer extensions starting with `OES_vertex_array_object`, should be done, right?

Palmer-JC:
Ok, I find how you are doing VertexArrays both ways, so problem is elsewhere

modulesio:
Yup the review is more for my own notes on what needs to be done before merging.

Vertex arrays should indeed have already be implemented, including the `OES_vertex_array_object` extension for WebGL backwards compatibility.

I'm not yet sure what's wrong with running that simple example though...

Palmer-JC:
Well spent too much time trying to figure out what is wrong with that page. Since in my version of the page, I have commented out any of the Query functions, this is an unrelated problem.  Maybe a separate issue for this.

Good news is I just found another [Examples for WebGL2 site](https://tsherif.github.io/webgl2examples/) / repo.  When you run the OcclusionQuery there on exokit, you get the same culling occurring in the bottom left.  See image.  When you run this on the desktop, you can turn off the query with a checkbox, then you always see all the balls.  So it is working, though I still have to verify that all of the calls are being used.

occusionquery

The last thing is it just screams on exokit compared to browsers.  This sort of triggered a memory that it might be being throttled.  Need to check khronos DL archives.  Think there might be some kind of weird hack they are thwarting.  If so, will try to register on DL to ask if this has any meaning for a ML type device, especially for an embedded technology.

Getting close!

Palmer-JC:
That conversation was about [EXT_disjoint_timer_query](https://developer.mozilla.org/en-US/docs/Web/API/EXT_disjoint_timer_query) & this [glitch](https://www.vusec.net/wp-content/uploads/2018/05/glitch.pdf) rowhammer attack.  Looks like that's what they called Query when it was an extension in WebGL1.

modulesio:
Makes sense, super interesting research! We don't do any sort of mitigations to protect from the app on the Exokit side, so that's probably the reason.

Palmer-JC:
Maybe adding a [command line](https://github.com/webmixedreality/webmr-docs/blob/master/docs/exokitCommandLineFlags.md) switch to allow query before going to production, or throw an error in create.  Could also try to slow down like others, but I do not know how to do that.

The advantage of a switch, is if making an embedded app, you do not get a penalty.

I only needed `if(!gl.deleteQuery)`, or the function to exist to BabylonJS

modulesio:
For exokit's case I think it's fine to just implement it as a direct binding. If we are trying to mitigate things like rowhammer there are many other things we'd need to do once we get in that boat -- so if anything that should be a separate issue.

Palmer-JC:
Ok, the example is using everything but `isQuery`, `getQuery`, & `deleteQuery`.  I added `isQuery` & `deleteQuery` and they work fine.

`getQuery` was also added, not getting the results I thought I was asking for.  It did wrapper the call though.  Calling it through Firefox & Chrome behaved the same way.  

The problem was that a constant required for the call is NOT in the spec!!  I added an [issue ](https://github.com/KhronosGroup/WebGL/issues/2809)on the WebGL repository.  This was F'ing painful to figure out.

I added the constant, and it now works.  There is just one thing left, if the parameter is wrong `getQuery` is supposed to return `null`.  What is that in `info` terms.? Here is current wrapper
```c++
NAN_METHOD(WebGL2RenderingContext::GetQuery) {
  GLenum target = info[0]->Int32Value();
  GLenum pname = info[1]->Int32Value();
  GLint value;
  
  glGetQueryiv(target, pname, &value);
  if (glGetError() == GL_NO_ERROR) {
    Local<Object> queryObject = Nan::New<Object>();
    queryObject->Set(JS_STR("id"), JS_INT(value));

    info.GetReturnValue().Set(queryObject);
  }
  else Nan::ThrowTypeError("oh crap what to do?");
}
```

modulesio:
> What is that in `info` terms.?
`Nan::Null()`, or...

```
info.GetReturnValue().Set(Nan::Null());
```

modulesio:
Re: the spec, I wonder if that is an intentional omission from WebGL2? Regardless, I don't think the answer affects this PR.

Palmer-JC:
The spec omission I figured out a few minutes after I posted it & closed that.  I was putting `GL_` on the front of the constant in the test, smack head.  

Probably never going to know why the exokit implementation is insanely faster than all others, but it works, and that is all that really counts.  I checked the flags.  The option is to un-capped, so capped must be the default.

I can re-organize / drop dups  the constants, if you want.

Palmer-JC:
The link in the transform feedback PR was a local file.  The web version is:
[https://tsherif.github.io/webgl2examples/particles.html](https://tsherif.github.io/webgl2examples/particles.html)

Palmer-JC:
Remainder of Sampler calls tested without changes required.  Probably done working in this area for now.  May look into WebAudio in the next few days.  Nothing works for me there.

modulesio:
Great, thanks. I'll give it a test.

modulesio:
We tested this and it works on the list at https://tsherif.github.io/webgl2examples/

Palmer-JC:
These can also probably be closed:
https://github.com/webmixedreality/exokit/issues/189
https://github.com/webmixedreality/exokit/issues/190
https://github.com/webmixedreality/exokit/issues/191

modulesio:
Thanks @Palmer-JC, good eye on those. I closed them out.

Issue #758, opened 2019-01-22
Reality tabs sticky shift


modulesio:
In the reality tabs example, in-XR, it's hard to type capitals since the shift key is not sticky. We should probably toggle it down for one character like mobile.

Tangential to this is having a functional caps lock on the virtual keyboard.

modulesio:
Closing because fronted work has moved to https://github.com/exokitxr/exokit-frontend.

PR #776, opened 2019-02-15
Basic shift key stickiness

shawwn:
Related to #758


examples/realitytabs.html
```diff
@@ -1036,7 +1037,9 @@
           code === 27 || // esc
           code === 91 // win
         ) {
-          // nothing
+          if (code === 16) { // shift
+            isShifted = !isShifted;
```
modulesio: This might work functionally, but it would be good to actually render blue shift key overtop or something to signal to the user the key is "on".

modulesio:
Closing due to being outdated; the frontend work is at https://github.com/exokitxr/exokit-frontend now.

PR #779, opened 2019-02-20
include servo vs chromium on android compile flag

ngokevin:
A small step. The included Servo won't work on Android due to being ML-specific.

### DIFF
--- a/deps/exokit-bindings/browser/include/browser-common.h
+++ b/deps/exokit-bindings/browser/include/browser-common.h
@@ -16,7 +16,7 @@
 #include <mutex>
 #include <functional>
 
-#ifndef LUMIN
+#if !defined(LUMIN) && !defined(__ANDROID__)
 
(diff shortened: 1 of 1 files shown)


avaer:
To clarify, it looks like this pr is switching to servo for Android build, just like ML. Which makes sense. Was that the intent?

Either way we are currently trying to switch to libcef for all cases, though we would need an arm64 build included for that, which is currently not the case.

ngokevin:
I think I'm mistaken and should just include libcef in my Makefile. So we should instead build libcef for arm64 (like in native-browser-deps)?

Or I might just try to stub this out entirely for now just to focus on getting Android working since I'm not needing a 2D view at the moment.

avaer:
Yeah  I would stub it out for now. Getting Android working might be easier than getting libcef working on Android.

ngokevin:
Thanks for the info!

PR #782, opened 2019-02-24
Fix Magic Leap keyboard events

brianpeiris:
Make keycode switch cases into actual cases instead of jump labels 😅 
Also fix misspelled keypress event name.

Exokit would just hard crash if you typed a key on a bluetooth keyboard.


modulesio:
Thanks! This was definitely busted...

PR #787, opened 2019-02-26
Oculus Desktop SDK support

dmarcos:
Set up dependencies, native bindings, initialize Oculus runtime and check for connected Rift.

Work in progress not ready for merge. I wanted to share early, collect feedback and make sure I'm not going down the wrong path. Not yet familiarized with Exokit code base. 

I want to render something on screen before merge.


modulesio:
Out of band we discussed the Nvidia integrated/discrete GPU problem that affected this PR. If we pre-create a context on the integrated GPU, then `LibOVR` cannot use it and creating the swap chain will fail.

One potential option for fixing this is compiling this into the binary:

```
extern "C" {
  _declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
}
```

However, that probably doesn't work if you intend to use stock `Node.js`.

See also: https://stackoverflow.com/questions/6036292/select-a-graphic-device-in-windows-opengl

modulesio:
I fixed up the blitting and view/projection matrices in https://github.com/webmixedreality/exokit/commit/641294ac5ade07916dc39020dc7da6baa6a5c66f, https://github.com/webmixedreality/exokit/commit/9b39b1e497dbe14dd239897ba01754ef6665a655.

Feel free to pick them.

modulesio:
This might also benefit from the high-resolution headset fix in https://github.com/webmixedreality/exokit/pull/829.

deps/oculus/src/ovrsession.cpp
```diff
@@ -0,0 +1,436 @@
...
+    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+    glEnable(GL_FRAMEBUFFER_SRGB);
+
+    ovr_CommitTextureSwapChain(session, eyes[eye].ColorTextureChain);
+    ovr_CommitTextureSwapChain(session, eyes[eye].DepthTextureChain);
+
+    glFramebufferTexture2D(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, 0, 0);
+    glFramebufferTexture2D(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D, 0, 0);
+
+    if (gl->HasTextureBinding(gl->activeTexture, GL_TEXTURE_2D)) {
```
modulesio: We are binding the framebuffer above, likely overwriting the user's current binding. Whenever we do something like this, we need to restore the user's binding back. That is what e.g. `gl->HasFramebufferBinding` is for.

src/index.js
```diff
@@ -1165,157 +1293,181 @@ const _startRenderLoop = () => {
...
+          localPositionArray3,   // hmd position
+          localQuaternionArray4, // hmd orientation
+          localFloat32Array,     // left eye view matrix
+          localFloat32Array2,    // left eye projection matrix
+          localFloat32Array3,    // right eye view matrix
+          localFloat32Array4     // right eye projection matrix
+        );
+        vrPresentState.hasPose = true;
+
+        xrState.positon = localPositionArray3;
```
modulesio: Typo?

src/index.js
```diff
@@ -471,12 +474,131 @@ class XRState {
...
+
+        // fps = VR_FPS;
+
+        const system = vrPresentState.system = vrPresentState.system || nativeBindings.oculusVr.Oculus_Init();
+        const lmContext = vrPresentState.lmContext || (nativeBindings.nativeLm && new nativeBindings.nativeLm());
+
+        system.SetupSwapChain(contexts[0]);
+        const {width: halfWidth, height} = system.GetRecommendedRenderTargetSize();
+
+        const maxTextureSize = 4096;
```
modulesio: I think this was updated upstream to a different behavior for openvr. We should probably be consistent both ways, so maybe this should be broken out into a helper function.

More generally, all of this bootstrapping seems like it could be parameterized by the hardware API and most of this code could be shared between the VR platforms.

src/index.js
```diff
@@ -471,12 +474,131 @@ class XRState {
 }
 const xrState = GlobalContext.xrState = new XRState();
 
-if (nativeBindings.nativeVr) {
+if (nativeBindings.oculusVr) {
```
modulesio: If I'm reading this right that means when you compile with OVR, OpenVR would be disabled? Is that intended?

src/index.js
```diff
@@ -471,12 +474,131 @@ class XRState {
...
+  nativeBindings.nativeVr.requestPresent = function (layers) {
```
modulesio: I think this is slightly confusing to hijack `nativeBindings.nativeVr` (which was previously OpenVR) to bring up the Oculus device. Maybe it would be better to have both drivers separate rather than decorating each other's functions?

src/Window.js
```diff
@@ -476,6 +477,9 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
       if (nativeVr && nativeVr.VR_IsHmdPresent()) {
         result.push(window[symbols.mrDisplaysSymbol].vrDisplay);
       }
+      if (oculusVr && oculusVr.Oculus_IsHmdPresent()) {
+        result.push(window[symbols.mrDisplaysSymbol].vrDisplay);
```
modulesio: If both are present that would mean we have the same VR display twice? It might make more sense to make the oculus device completely separate from the OpenVR one.

deps/oculus/src/ovrsession.cpp
```diff
@@ -0,0 +1,404 @@
...
+    Local<String> width_prop = Nan::New<String>("width").ToLocalChecked();
+    Nan::Set(result, width_prop, Nan::New<Number>(width));
+
+    Local<String> height_prop = Nan::New<String>("height").ToLocalChecked();
+    Nan::Set(result, height_prop, Nan::New<Number>(height));
+  }
+  info.GetReturnValue().Set(result);
+}
+
+NAN_METHOD(OVRSession::RequestGetPoses) {
```
modulesio: One thing to keep in mind is that the OpenVR and magic leap bindings [are multi-threaded](https://github.com/webmixedreality/exokit/blob/30921aaf7197defbd8463e88cf1f8f0ff6ca5861/deps/openvr/src/ivrcompositor.cpp#L222), and I believe this one is not.

This means that for Oculus, this binding will hold up the Node render loop and things like network requests would be frame-locked. That's not a bug but it would probably be a notable experience degradation. Additionally, if we merge parallel user threads (https://github.com/webmixedreality/exokit/pull/760), the single-thread solution cannot work at all since it relies on asynchrony of all threads.

The fix for this would be to make sure the pose wait and submit methods are asynchronous and return a promise, while runnig on another thread. The rest of the code would be the same. We don't need to do this before merging, but we'd need to do this soon either way.

src/Window.js
dmarcos: I changed this. If an Oculus headset is detected it takes precedence over OpenVR

src/index.js
dmarcos: OpenVR and Oculus headsets are now decoupled.

src/index.js
dmarcos: I don't think this is a typo. Can you elaborate?

modulesio: posit`I`on?

src/index.js
```diff
@@ -1089,9 +1211,14 @@ const _startRenderLoop = () => {
               nativeWindow.blitFrameBuffer(context, vrPresentState.msFbo, vrPresentState.fbo, vrPresentState.glContext.canvas.width, vrPresentState.glContext.canvas.height, vrPresentState.glContext.canvas.width, vrPresentState.glContext.canvas.height, true, false, false);
             }
 
-            vrPresentState.compositor.Submit(context, vrPresentState.tex);
-            vrPresentState.hasPose = false;
+            if (vrPresentState.oculusSystem) {
+              nativeBindings.nativeWindow.setCurrentWindowContext(windowHandle);
+              vrPresentState.oculusSystem.Submit(context, vrPresentState.tex, vrPresentState.fbo, vrPresentState.glContext.canvas.width, vrPresentState.glContext.canvas.height);
+            } else {
```
modulesio: Maybe it should assume less?

```
} else if (vrPresentState.compositor) {
```

src/index.js
```diff
@@ -1177,157 +1304,181 @@ const _startRenderLoop = () => {
...
+        xrState.rightViewMatrix.set(localFloat32Array3);
+        xrState.rightProjectionMatrix.set(localFloat32Array4);
+
+        localVector.toArray(xrState.position);
+        localQuaternion.toArray(xrState.orientation);
+        // console.log("CACA " +
+        //   xrState.positon[0] + ' ' +
+        //   xrState.positon[1] + ' ' +
+        //   xrState.positon[2] + ' '  );
+      } else {
```
modulesio: I would check that we're actually running an openVR loop (`vrPresentState.compositor`) rather than defaulting to that assumption.

src/index.js
dmarcos: good catch

modulesio:
[Here](https://github.com/webmixedreality/exokit/commit/ffa8ab618c6c939f84d6fe7cc28458ea258b1c9d) I added a few fixes to WebXR necessary for the rename. Still not rendering to OpenVR yet through.

dmarcos:
> 
> 
> [Here](https://github.com/webmixedreality/exokit/compare/display-rn-bugfix) I added a few fixes to WebXR necessary for the rename. Still not rendering to OpenVR yet through.

Thanks. Just cherry-picked the commit.

src/Window.js
```diff
@@ -475,9 +476,18 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
       if (nativeMl && nativeMl.IsPresent()) {
         result.push(window[symbols.mrDisplaysSymbol].mlDisplay);
       }
-      if (nativeVr && nativeVr.VR_IsHmdPresent()) {
-        result.push(window[symbols.mrDisplaysSymbol].vrDisplay);
+
+      // Oculus runtime takes precedence over OpenVR for Oculus headsets.
+      if (nativeOculusVR && nativeOculusVR.Oculus_IsHmdPresent()) {
+        GlobalContext.isOculusRuntime = true;
```
modulesio: I think thins might result in race conditions, where depending on whether you get the display or the controllers first, the controllers will be called something different. Additionally I don't think a global should be set from a getter.

Maybe the better approach would be to bootstrap controllers in this flow as well?

deps/oculus/src/ovrsession.cpp
```diff
@@ -0,0 +1,456 @@
...
+    Local<String> width_prop = Nan::New<String>("width").ToLocalChecked();
+    Nan::Set(result, width_prop, Nan::New<Number>(width));
+
+    Local<String> height_prop = Nan::New<String>("height").ToLocalChecked();
+    Nan::Set(result, height_prop, Nan::New<Number>(height));
+  }
+  info.GetReturnValue().Set(result);
+}
+
+NAN_METHOD(OVRSession::GetPose) {
```
modulesio: I think this will need to be run in a thread soon -- either in this PR or it will block other PRs.

I think the `GetPose`/`Submit` flow is throttled by the oculus runtime to maintain framerate -- by design -- but this will probably lock up the node loop while it's happening, resulting in things like network not processing, or being reduced in throughput.

The strategy elsewhere is to do this work in a thread/Promise wrapping, while leaving the main loop to continue working while the runtime is waiting for the headset to flip framebuffers.

Also I think this is a logical conflict with the threading PR and might result in total lockups there (https://github.com/webmixedreality/exokit/pull/760).

modulesio: Here is the above strategy executed for the OpenVR case:

https://github.com/webmixedreality/exokit/blob/607d5a9512fe79df7a56551325065e62d24934a1/src/index.js#L1160

https://github.com/webmixedreality/exokit/blob/607d5a9512fe79df7a56551325065e62d24934a1/deps/openvr/src/ivrcompositor.cpp#L280

src/Window.js
dmarcos: I think if we follow spec / other browsers implementation the problem cannot happen because VR controllers only show up in the gamepad API after starting to present.

dmarcos: I added a check to return empty array and not initialize gamepads if we're not presenting https://github.com/webmixedreality/exokit/pull/787/commits/600e9c254286e11f7edcd0ca534f35d213497a18

modulesio: > I think if we follow spec / other browsers implementation the problem cannot happen because VR controllers only show up in the gamepad API after starting to present.

Ok, that makes sense.

Is https://github.com/webmixedreality/exokit/commit/600e9c254286e11f7edcd0ca534f35d213497a18 the right commit?  I don't see any gamepads code there.

modulesio: In that case, should we move the check to the Gamepad getter entirely? If the intent is to return certain controllers when oculus is presenting, we can probably check whether the devices are actually presenting (https://github.com/webmixedreality/exokit/pull/787#diff-3db2392f75003eacb1df3b866c44fb3cR1362) rather than adding another flag to track it on display lookup time.

deps/oculus/src/ovrsession.cpp
```diff
@@ -1,14 +1,46 @@
...
+  std::deque<std::function<void()>> reqCbs;
+  std::deque<std::function<void()>> resCbs;
+  std::thread reqThead;
+
+  void RunResInMainThread(uv_async_t *handle) {
+    Nan::HandleScope scope;
+
+    std::function<void()> resCb;
+    {
+      std::lock_guard<std::mutex> lock(reqMutex);
```
modulesio: Should this be `resMutex`? It could result in corruption if we're not locking the right thing.

modulesio:
I tried this branch with a Vive connected and it loaded the Oculus software without presenting to vive:

```node --experimental-worker . examples\hello_xr.html
THREE.WebGLRenderer 96dev
request device
02/04 09:37:33.137 {INFO}    [Kernel:Default] [CAPI] LibOVR module is located at C:\Program Files\Oculus\Support\oculus-runtime\LibOVRRT64_1.dll
02/04 09:37:33.140 {INFO}    [Client] Connected to the server running version (prod = 1).1.36.0(build = 215623) feature version = 0. Client runs version (prod = 1).1.36.0(build = 0) feature version = 0
request session
request first frame
{ Error: Hmd Not Found (108)
    at Object.nativeBindings.nativeOpenVR.requestPresent (C:\Users\avaer\Documents\GitHub\exokit\src\index.js:603:77)
    at XRDevice.openVRDevice.onrequestpresent.layers [as onrequestpresent] (C:\Users\avaer\Documents\GitHub\exokit\src\Window.js:1336:60)
    at new XRWebGLLayer (C:\Users\avaer\Documents\GitHub\exokit\src\XR.js:338:22)
    at WebXRManager.setSession (three.js:21642:25)
    at session.requestAnimationFrame (file:///C:/Users/avaer/Documents/GitHub/exokit/examples/hello_xr.html:127:21)
    at animationFrame.device.onrequestanimationframe.timestamp (C:\Users\avaer\Documents\GitHub\exokit\src\XR.js:125:9)
    at tickAnimationFrame (C:\Users\avaer\Documents\GitHub\exokit\src\Window.js:1264:15)
    at Immediate._renderLoop [as _onImmediate] (C:\Users\avaer\Documents\GitHub\exokit\src\index.js:1767:18)
    at processImmediate (timers.js:632:19)
    at process.topLevelDomainCallback (domain.js:120:23) code: 108 }
THREE.FBXLoader: FBX binary version: 740```

File: deps/oculus/src/ovrsession.cpp [hunk 13]
```diff
@@ -0,0 +1,507 @@
...
+      ObjectWrap::Unwrap<OVRSession>(info.Holder())->ResetSession();
+      session = *ObjectWrap::Unwrap<OVRSession>(info.Holder())->session;
+      ObjectWrap::Unwrap<OVRSession>(info.Holder())->hmdMounted = true;
+    }
+  } else {
+    ObjectWrap::Unwrap<OVRSession>(info.Holder())->hmdMounted = false;
+  }
+
+  WebGLRenderingContext *gl = node::ObjectWrap::Unwrap<WebGLRenderingContext>(Local<Object>::Cast(info[0]));
+  GLuint textureSource = info[1]->Uint32Value();
```
modulesio: I think this is unused, since it's part of `fboIdSource`.

deps/oculus/src/ovrsession.cpp
```diff
@@ -0,0 +1,507 @@
...
+      0, 0,
+      eyes[eye].textureSize.w, eyes[eye].textureSize.h,
+      GL_COLOR_BUFFER_BIT,
+      GL_LINEAR
+    );
+
+    ovr_CommitTextureSwapChain(session, eyes[eye].ColorTextureChain);
+    ovr_CommitTextureSwapChain(session, eyes[eye].DepthTextureChain);
+
+    if (gl->HasFramebufferBinding(GL_READ_FRAMEBUFFER)) {
```
modulesio: This can be move out of the eye loop to eliminate extra rebindings per frame.

deps/oculus/src/ovrsession.cpp
```diff
@@ -0,0 +1,507 @@
...
+  // Declare human-readable name for this wrapper.
+  tpl->SetClassName(Nan::New("OVRSession").ToLocalChecked());
+
+  // Declare the stored number of fields (just the wrapped C++ object).
+  tpl->InstanceTemplate()->SetInternalFieldCount(1);
+
+  Nan::SetPrototypeMethod(tpl, "GetControllersInputState", GetControllersInputState);
+  Nan::SetPrototypeMethod(tpl, "GetPose", GetPose);
+  Nan::SetPrototypeMethod(tpl, "Submit", Submit);
+  Nan::SetPrototypeMethod(tpl, "SetupSwapChain", SetupSwapChain);
```
modulesio: From my reading this isn't called from JS and swap chains are managed from C++. Maybe we should get rid of this exposure and the `NAN_METHOD` to prevent confusion?

modulesio: Actually, I just saw [this](https://github.com/exokitxr/exokit/pull/787/files#diff-1fdf421c05c1140f6d71444ea2b27638R484). This is confusing since it seems both JS and C++ manage the swap chain.

I think if that's the case we should settle on one code path to do it -- in this case it seems `Oculus_Init` in C++ could set up the first swap chain.

src/VR.js
```diff
@@ -306,7 +308,7 @@ class FakeVRDisplay extends VRDisplay {
   constructor(window) {
     super('FAKE');
 
-    this.window = window;
+    this.window = globalWindow = window;
```
modulesio: I don't think `globalWindow` passed here is a valid approach; `Gamepad`s can be created from many different contexts, perhaps with no `window` global at all.

If the intent is to bind it for the sake of `getGamepads`, we can pass that in as a closure on `getGamepads` from wherever `navigator` is created (`Window.js`).

deps/oculus/include/ovrsession.h
```diff
@@ -0,0 +1,188 @@
...
+
+  // Node construction method for new instances.
+  static NAN_METHOD(New);
+  static NAN_METHOD(GetControllersInputState);
+  static NAN_METHOD(GetPose);
+  static NAN_METHOD(Submit);
+  static NAN_METHOD(GetRecommendedRenderTargetSize);
+
+  /// Create a singleton reference to a constructor function.
+  static inline Nan::Persistent<v8::Function>& constructor()
```
modulesio: This is more a suggestion, but I would keep implementations in the c++ side of this file, for compilation performance and code duplication reasons.

If we include this file from two objects, the linker might reject it as an error, which would be surprising and confusing.

src/VR.js
```diff
@@ -605,19 +604,30 @@ class FakeVRDisplay extends VRDisplay {
 
 const createVRDisplay = () => new FakeVRDisplay();
 
-const getGamepads = (() => {
+const getGamepads = function(window) {
+
+  let globalWindow = window;
```
modulesio: Is there a stylistic reason for renaming window here?

deps/oculus/include/ovrsession.h
dmarcos: I've split declaration / definition

deps/oculus/include/ovrsession.h
```diff
@@ -0,0 +1,188 @@
...
+  {
+    static Nan::Persistent<v8::Function> the_constructor;
+    return the_constructor;
+  }
+
+  static void ovrLogCallback(uintptr_t userData, int level, const char *message) {
```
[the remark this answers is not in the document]
dmarcos: Renamed

src/VR.js
dmarcos: `window` is a pretty overloaded variable name. Intent was to convey that this is something different than your usual global `window` reference.

dmarcos: renamed

PR #792, opened 2019-03-04
Add --help

shawwn:


modulesio:
This is a good idea, but is there some way to autogenerate this so it doesn't get out of sync with the actual arguments parsing?

shawwn:
Not without reworking how the argument system works.

modulesio:
Yes, I am recommending a slight refactor of that code for the sake of maintainability.

modulesio:
This is not the right strategy for a `--help` documentation.

We might derive one from the docs at some point.

PR #793, opened 2019-03-04
Fix --require

shawwn:


src/Window.js
```diff
@@ -388,7 +388,7 @@ const _makeWindow = (options = {}, parent = null, top = null) => {
   utils._storeOriginalWindowPrototypes(window, symbols.prototypesSymbol);
 
   const windowStartScript = `(() => {
-    ${!options.args.require ? 'global.require = undefined;' : ''}
+    ${(!GlobalContext.args.require && !options.noRequire) ? 'global.require = undefined;' : ''}
```
modulesio: Does `options.noRequire` exist?

shawwn: No, but I thought it might be a good idea to let users of the `Window` object have the option of ignoring the global `--require` option. But exokit is the only user of this code right now, so maybe it doesn't matter.

modulesio: Yeah, I would delete the nonexistent option for now.

modulesio:
Fixed in `user-parallel`.

PR #794, opened 2019-03-04
Add --no-bugsnag

shawwn:
Bugsnag was prematurely terminating exokit when loading from localhost:8000/tests/test_celeste.html (presumably because my test has errors, which is expected behavior) so I added `--no-bugsnag` to prevent exokit from quitting.

### DIFF
--- a/src/bugsnag.js
+++ b/src/bugsnag.js
@@ -19,7 +19,7 @@ const bugsnagApiKey = (() => {
     return null;
   }
 })();
-if (bugsnagApiKey) {
+if (bugsnagApiKey && !GlobalContext.args.noBugsnag) {
   bugsnag.register(bugsnagApiKey, {
     metaData: {
       argv: process.argv,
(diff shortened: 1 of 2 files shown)


modulesio:
Can we fix the bugsnag case so that exokit continues, rather than having this flag?

shawwn:
Eventually. But this flag is good to have regardless. Bugsnag has caused problems to end-users (me) on several occasions, and it's not very practical to force users to drop everything and fix a bugsnag issue just to get their demo to run.

modulesio:
That sounds like a good reason to prioritize fixing the bug. I don’t think lack of understanding is a good reason to add a new flag feature.

So you have a small repro so this can very converted into a concrete issue so we can fix the root cause?

modulesio:
Bugsnag/native stack capture is [currently disabled](https://github.com/exokitxr/exokit/blob/764f44e3410d3afe15f4747478ebd2ff8f8bd322/src/index.js#L1270), since it interfered with the move to Node 12.

Closing this issue as effectively fixed.

PR #795, opened 2019-03-04
Fix window.location.href = "..."

### LINKED ISSUE #778: ML1: window.location.href = ... throws error

shawwn:
Closes #778

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -55,6 +55,7 @@ nativeBindings.nativeWindow.setEventLoop(eventLoopNative);
 
 const GlobalContext = require('./GlobalContext');
 GlobalContext.commands = [];
+GlobalContext.exokit = core;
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -55,6 +55,7 @@ nativeBindings.nativeWindow.setEventLoop(eventLoopNative);
 
 const GlobalContext = require('./GlobalContext');
 GlobalContext.commands = [];
+GlobalContext.exokit = core;
```
modulesio: Instead of globally exposing, is there a way we could require the core instead?

shawwn: Note that attempting to require the core from within webworkers seems to explode. It may be a better idea to expose this globally. Not sure yet.

shawwn:
Closing in favor of #847

PR #796, opened 2019-03-04
Implement document.queryCommandSupported

shawwn:


modulesio:
I don’t think this is right, since some of the copy-paste commands are supported.

What are the commands required to get this to true? Can we enumerate those?

PR #799, opened 2019-03-04
Shim window.matchMedia method

shawwn:
It'd take some work to implement https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia correctly, but this seems to let sites at least not fail with an error.

### DIFF
--- a/src/Window.js
+++ b/src/Window.js
@@ -326,6 +326,9 @@ class Worker {
   set onerror(onerror) {
     this.worker.onerror = onerror;
   }
+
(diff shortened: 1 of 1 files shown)


modulesio:
I would rather wait for an actual implementation.

Additionally, I don't think `false` return is legal, and this PR also adds an unintended `terminate` method.

modulesio:
Closing in favor of the more complete https://github.com/exokitxr/exokit/pulls.

PR #800, opened 2019-03-04
Fix chdir error in webworkers

shawwn:
Possibly a dumb fix, since the following `require` statement also fails with "Module failed to self-register" error. But I'll tackle that next.


modulesio:
This should all go away with the introduction of parallel windows and workers. If the native module require doesn’t work then I don’t think this PR adds anything yet.

modulesio:
Multithreading shipped, so this issue does not apply anymore.

PR #803, opened 2019-03-04
Add initial crypto.subtle polyfill

modulesio:


RangerMauve:
I suggest adding [this implementation](https://github.com/PeculiarVentures/webcrypto) as a dependency.

modulesio:
Thanks @RangerMauve, I'll break that out to a separate issue.

PR #847, opened 2019-03-11
Fix assignment to window.location.href

### LINKED ISSUE #778: ML1: window.location.href = ... throws error

shawwn:
Closes #778

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -315,6 +315,7 @@ const exokit = (s = '', options = {}) => {
   options.dataPath = options.dataPath || __dirname;
   options.args = options.args || {};
   options.replacements = options.replacements || {};
+  options.loadCallback = options.loadCallback || ((href, opts) => exokit.load(href, opts));
   return _makeWindowWithDocument(s, options);
 };
 exokit.load = (src, options = {}) => {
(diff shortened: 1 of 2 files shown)


src/core.js
```diff
@@ -315,6 +315,7 @@ const exokit = (s = '', options = {}) => {
   options.dataPath = options.dataPath || __dirname;
   options.args = options.args || {};
   options.replacements = options.replacements || {};
+  options.loadCallback = options.loadCallback || ((href, opts) => exokit.load(href, opts));
```
modulesio: In general, we want `options` to be serializable (and this is a requirement for `user-parallel` in which `window`s do not share any namespace or closures). So I don't think this is an architecture that would work going forward.

modulesio:
I still think this is an important PR to get in; can we architect this flow so that the location update event bubbles up to either `index.js` or `core.js`, where the load can be effected?

In the `user-parallel` branch there is a `WindowVm` parent owner of all windows, so this approach would translate over there.

modulesio:
`window.location.href` assignment works completely different now that we have isolated multithreading; this PR no longer applies.

PR #848, opened 2019-03-11
Update WebGLRenderingContext::GetParameter

shawwn:
This PR fixes various GetParameter results so that the following demo runs (mostly) identically in Chrome and Exokit:

https://github.com/SaschaWillems/webgl/blob/master/webgl2_capsviewer/index.html

--

Side note: the demo tries to do this:

```
if (params.constructor === Int32Array) {
```

... which returns true in Chrome but false in Exokit.

Maybe _normalizePrototype needs to be called on GetParameter's results?

### DIFF
--- a/deps/exokit-bindings/webglcontext/src/webgl.cc
+++ b/deps/exokit-bindings/webglcontext/src/webgl.cc
@@ -4440,7 +4440,6 @@ NAN_METHOD(WebGLRenderingContext::GetParameter) {
     case GL_MAX_COMBINED_FRAGMENT_UNIFORM_COMPONENTS:
     case GL_MAX_COMBINED_UNIFORM_BLOCKS:
     case GL_MAX_COMBINED_VERTEX_UNIFORM_COMPONENTS:
-    case GL_MAX_ELEMENT_INDEX:
     case GL_MAX_ELEMENTS_INDICES:
     case GL_MAX_ELEMENTS_VERTICES:
     case GL_MAX_FRAGMENT_INPUT_COMPONENTS:
@@ -4448,14 +4447,12 @@ NAN_METHOD(WebGLRenderingContext::GetParameter) {
     case GL_MAX_FRAGMENT_UNIFORM_COMPONENTS:
     case GL_MAX_PROGRAM_TEXEL_OFFSET:
     case GL_MAX_SAMPLES:
-    case GL_MAX_SERVER_WAIT_TIMEOUT:
     case GL_MAX_TEXTURE_LOD_BIAS:
     case GL_MAX_TRANSFORM_FEEDBACK_INTERLEAVED_COMPONENTS:
     case GL_MAX_TRANSFORM_FEEDBACK_SEPARATE_ATTRIBS:
     case GL_MAX_TRANSFORM_FEEDBACK_SEPARATE_COMPONENTS:
     case GL_MAX_UNIFORM_BLOCK_SIZE:
     case GL_MAX_UNIFORM_BUFFER_BINDINGS:
-    case GL_MAX_VARYING_COMPONENTS:
     case GL_MAX_VERTEX_OUTPUT_COMPONENTS:
     case GL_MAX_VERTEX_UNIFORM_BLOCKS:
     case GL_MAX_VERTEX_UNIFORM_COMPONENTS:
@@ -4465,6 +4462,18 @@ NAN_METHOD(WebGLRenderingContext::GetParameter) {
       GLint param;
       glGetIntegerv(name, &param);
       info.GetReturnValue().Set(JS_INT(param));
+      break;
+    }
+    case GL_MAX_VARYING_COMPONENTS:
+    {
+      // return an int
+      GLint param;
+      glGetIntegerv(GL_MAX_VARYING_VECTORS, &param);
+      if (param >= 0) {
+        info.GetReturnValue().Set(JS_INT(4*param));
+      } else {
+        info.GetReturnValue().Set(JS_INT(param));
+      }
       break;
     }
     case GL_DEPTH_CLEAR_VALUE:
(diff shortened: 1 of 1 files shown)


deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4467,6 +4464,18 @@ NAN_METHOD(WebGLRenderingContext::GetParameter) {
       info.GetReturnValue().Set(JS_INT(param));
       break;
     }
+    case GL_MAX_VARYING_COMPONENTS:
```
shawwn: `gl.getParameter(gl.MAX_VARYING_COMPONENTS)` was giving an error, but according to the spec `MAX_VARYING_COMPONENTS` is always identical to `4*MAX_VARYING_VECTORS`, and `gl.getParameter(gl.MAX_VARYING_VECTORS)` seems to work fine.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4511,11 +4520,7 @@ NAN_METHOD(WebGLRenderingContext::GetParameter) {
       // return a int32[2]
       GLint params[2];
       glGetIntegerv(name, params);
-
-      Local<Array> arr = Nan::New<Array>(2);
-      arr->Set(0,JS_INT(params[0]));
-      arr->Set(1,JS_INT(params[1]));
-      info.GetReturnValue().Set(arr);
+      info.GetReturnValue().Set(createTypedArray<Int32Array>(2, params));
```
shawwn: (gl.getParameter returns typed arrays in Chrome.)

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4604,14 +4593,10 @@ NAN_METHOD(WebGLRenderingContext::GetParameter) {
...
 
-      unique_ptr<GLint[]> params(new GLint[numFormats]);
-      glGetIntegerv(name, params.get());
-
-      Local<Array> arr = Nan::New<Array>(numFormats);
-      for (size_t i = 0; i < numFormats; i++) {
-        arr->Set(i, JS_INT(params[i]));
-      }
-      info.GetReturnValue().Set(arr);
+      GLint* params = new GLint[numFormats];
```
shawwn: note that numFormats could be -1 here if glGetIntegerv returns an error (but this was true of the old code as well)

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4630,6 +4615,22 @@ NAN_METHOD(WebGLRenderingContext::GetParameter) {
       info.GetReturnValue().Set(JS_BOOL(gl->premultiplyAlpha));
       break;
     }
+    case MAX_CLIENT_WAIT_TIMEOUT_WEBGL:
+    case GL_MAX_SERVER_WAIT_TIMEOUT:
+    {
```
shawwn: These were returning -1 in exokit but 0 in Chrome. Since they're timeouts, is it safe to punt here and always return 0?

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4630,6 +4615,22 @@ NAN_METHOD(WebGLRenderingContext::GetParameter) {
...
+      info.GetReturnValue().Set(JS_INT(0));
+      break;
+    }
+    case GL_MAX_ELEMENT_INDEX:
```
shawwn: `gl.getParameter(gl.MAX_ELEMENT_INDEX)` was returning -1 but also causing a gl error. I'm not sure why.

Chrome gives `0xffffffff` i.e. `(GLuint)-1`, and this code seems to match Chrome now.

(Why does Chrome work fine?)

modulesio:
Thanks for this!

> Maybe _normalizePrototype needs to be called on GetParameter's results?

Indeed, this would be true until `user-parallel` lands.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4630,6 +4615,22 @@ NAN_METHOD(WebGLRenderingContext::GetParameter) {
...
+    {
+      GLint64 elementIndex = -1;
+      glGetInteger64v(GL_MAX_ELEMENT_INDEX, &elementIndex);
+      if (glGetError() == GL_NO_ERROR) {
```
modulesio: I am wary of doing any `glGetError` ourselves, because this could stomp on _actual_ unrelated gl errors that the user would want to see. If we capture the error here, and we hit that corner case, this could result in very mysterious and hard-to-debug bugs seen by the user.

Additionally, I think `glGetError` might force a flush of the pipeline, making this a stalling GL call when it might not be otherwise.

Is there some other way to do this without eating the error?

shawwn: agreed.

one possibility: we could skip calling `glGet*` and just return  `0xffffffff` (i.e. just lie).

Here's all that the spec says:

```
GL_MAX_ELEMENT_INDEX
data returns a single value, the maximum index that may be specified during the transfer of generic vertex attributes to the GL.

GL_MAX_ELEMENT_INDEX is accepted by pname only if the GL version is 4.3 or greater.
```
I don't know why it works in Chrome but not Exokit.

modulesio: > GL_MAX_ELEMENT_INDEX is accepted by pname only if the GL version is 4.3 or greater.

This might be it. We are [GL 3.2](https://github.com/webmixedreality/exokit/blob/88d5da063d08959a7b8c314dad61591523116c87/deps/exokit-bindings/glfw/src/glfw.cc#L1131), I think mostly because macOS.

deps/exokit-bindings/webglcontext/src/webgl.cc
modulesio: Good catch.

deps/exokit-bindings/webglcontext/src/webgl.cc
modulesio: Curious if this was because of a legitimate GL error...

deps/exokit-bindings/webglcontext/src/webgl.cc
modulesio: Would that ever be a legit value, or only if we have a broken build/driver?

modulesio:
@shawwn Can we update this PR to get it into a mergeable state?

deps/exokit-bindings/webglcontext/src/webgl.cc
modulesio: Which operating system/GPU driver was this tested on? If the GL spec says `GL_MAX_VARYING_COMPONENTS` exists, then we should probably query that.

modulesio:
Closing in favor of https://github.com/exokitxr/exokit/pull/941.

PR #856, opened 2019-03-15
Add Progressive Web App examples

chrislatorres:
This PR adds per-example manifests and service worker scripts. The goal is to implement in Exokit so that: 

The PWA links to manifest and registers the service worker in its HTML, the service worker script reads in the set of files for the app and caches it up, the resulting "cache" is the set of files of the app. 

The entire app can now be "offline" as a PWA, and now you have your full app that can be packaged, into MPK or otherwise. :thumbsup:


chrislatorres:
What are you thinking for the clean up?

- Per-example directory
- Assets directory for png, jpg, glb
- `script` directory for non example-specific .js files

The above would be a lot of changes though which might be better in its own issue

modulesio:
Yeah, I was just thinking the above. It could be a separate issue.

PR #860, opened 2019-03-16
Frame reprojection example

modulesio:
Adds a frame reprojection shader example to `examples`. The intent is for this to be a testbed for actual frame reprojection support in exokit code.

The main use case for this is the scenario where many reality tabs are composited and not all can keep up. In those cases, we can reproject older reality tab framebuffers (with depth) so that the fast reality tabs are not held up.

This is a universal feature of VR platforms, but since in this instance we _are_ the VR platform and we wrote out own timing and compositor, we need to also handle the reprojection frame-perfectly.

The style of this reprojection shader is the one used on recent versions of OpenVR, where we re-swim the depthed canvas in front of the eye with a high resolution geometry mesh on the framebuffer.

### DIFF
--- a/examples/OrbitControls.js
+++ b/examples/OrbitControls.js
@@ -1,0 +1,1070 @@
+/**
+ * @author qiao / https://github.com/qiao
+ * @author mrdoob / http://mrdoob.com
+ * @author alteredq / http://alteredqualia.com/
+ * @author WestLangley / http://github.com/WestLangley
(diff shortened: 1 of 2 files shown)


modulesio:
Got rid of some of the glitching here. This now handles the case where the "tail" of an AR-type render gets reprojected out to infinity depth. We simply discard fragments in which the reprojection causes more than a small change in Z.

PR #872, opened 2019-03-19
Emit vrdisplaychange only when the presenting state of a VR display changes

dmarcos:
As per spec. Event shouldn't fire on every requestPresent / exitPresent call:

https://developer.mozilla.org/en-US/docs/Web/API/Window/vrdisplaypresentchange_event

### DIFF
--- a/src/VR.js
+++ b/src/VR.js
@@ -183,7 +183,7 @@ class VRStageParameters {
 class VRDisplay extends EventEmitter {
   constructor(displayName) {
     super();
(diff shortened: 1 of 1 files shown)


modulesio:
From MDN:

> The vrdisplaypresentchange event of the WebVR API is fired when the presenting state of a VR display changes — i.e. goes from presenting to not presenting, or vice versa.

In which case indeed, the old code would be in error.

PR #874, opened 2019-03-20
Emit vrdisplaypresentchange only when VRDisplay changes presentation state

dmarcos:


modulesio:
Previous PR only covered the fake vr display: https://github.com/webmixedreality/exokit/pull/872

PR #877, opened 2019-03-23
Node 12 jitless build

modulesio:
This bumps the build process to node 12 (compatibility was added in https://github.com/webmixedreality/exokit/pull/873).

- `libnode.a` is updated to node 12 `-pre` with V8 `7.4` via https://github.com/nodejs/node/pull/26685.
- build process uses local node 12 checkout for now to ensure proper sourcing of header files; this will not be needed at all when node 12 is released in Apr 2019

This allows running without executable memory on platforms that do not allow it, including iOS.

### DIFF
--- a/metadata/program-device.mabu
+++ b/metadata/program-device.mabu
@@ -19,7 +19,7 @@ SRCS = ../main.cpp
 
 # Include paths for any language are specified here.
 #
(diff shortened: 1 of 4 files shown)


modulesio:
Support is in user-parallel, but it mostly won’t matter until we ship iOS support.

PR #879, opened 2019-03-26
ServiceWorker support

modulesio:
This PR adds support to Exokit for [Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API), which is an API for letting the app control its own network behavior, including programmable caching.

The end goal is full progressive web app support for Exokit, allowing applications to package themselves for offline usage (resulting in APK/MPK/exe).

This feature is dependent on [parallel window/worker threading](https://github.com/webmixedreality/exokit/pull/760).


modulesio:
Required intercept support in the network modules:

https://github.com/modulesio/window-xhr/pull/3
https://github.com/modulesio/window-fetch/pull/2

modulesio:
For this, exokit would need to implement the interceptors in these modules:

- https://github.com/modulesio/window-xhr/pull/3/files#diff-50e3aa130a4f97a42ee2cf111c7b1d9dR98
- https://github.com/modulesio/window-fetch/pull/2/files#diff-1fdf421c05c1140f6d71444ea2b27638R53

PR #887, opened 2019-04-03
Cleanup unused asset files and move README gifs to GitHub's CDN

chrisplatorres:
This PR cleans up unused asset files and moves the README gifs to [GitHub's CDN](https://gist.github.com/vinkla/dca76249ba6b73c5dd66a4e986df4c8d) to require less storage space for local repos.


modulesio:
Will the images get blown away in future from the README.md or does Github commit to the hosting forever?

chrisplatorres:
I couldn't find any exact info, I assume it will not get blown away because it is the CDN used for github issues.

modulesio:
I would guess there's no guarantee in that case and it could be sweeped by
a task in the future. I don't understand the assets moving part of this PR,
since it doesn't actually reduce repo size, and makes things less
convenient to update.

chrisplatorres:
> I would guess there's no guarantee in that case and it could be sweeped by a task in the future.

That is an issue with relying on CDNs in general, using the GitHub CDN is relatively good since it is relying on the same site. 

Worst case is if it is randomly sweeped, the gifs are recovered from history or remade. 

This PR does not move any assets, just deletes and moves the gifs to GitHub's CDN, which should save 120mb+ of storage space for local repos.

modulesio:
Thanks for the explanation, this LGTM.

PR #889, opened 2019-04-03
Move root cpp files to cpp directory

chrisplatorres:
This PR is to move the root .cpp files to a new `cpp` directory and update associated paths.


modulesio:
Should just make sure that magic leap continues to build.

modulesio:
Re: main.cpp, I think a better name for it would be `android.cpp` or `mobile.cpp` or `stub.cpp` or something. It's only purpose is to be a mobile bootstrap for `node`, and in fact is not used at all on desktop. So it could be confusing for newcomers to the codebase.

chrisplatorres:
Closing in favor of #989

PR #896, opened 2019-04-05
Add OpenVR trackers support

modulesio:
Adds OpenVR tracker support, exposing trackers as navigator `Gamepad` objects.

Fixes https://github.com/webmixedreality/exokit/issues/884.


modulesio:
This should be working now; up to 8 OpenVR trackers are supported with this PR.

There is a [`trackers.html` example](https://github.com/webmixedreality/exokit/pull/896/files#diff-bd3c61b6a5b33a94244fdb0700f4452c) which loads up all controllers and all trackers for testing.

PR #908, opened 2019-04-13
Update exokit.iss publisher name and copyright

chrisplatorres:
Noticed the windows installer publisher is "Zeo Incorporated", can't find any reference to zeo in the repo though, maybe it is something in the CI..

Looking in the installer config, the publisher in the config is WebMR, which has since been updated to exokitXR and the copyright was outdated.

zeoincwindows


modulesio:
The "Zeo Incorporated" refers to the certificate we're using, which would need to be re-purchased under a new name to change this.

However, I think otherwise this is a good change. 👍

PR #914, opened 2019-04-13
Fix 100% CPU usage

### LINKED ISSUE #907: Exokit CLI uses 100% CPU
**To Reproduce**
1. `node .`
2. observe 100% cpu usage

shawwn:
Closes #907

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1128,6 +1128,18 @@ const _startRenderLoop = () => {
   const TIMESTAMP_FRAMES = 100;
   // let lastFrameTime = Date.now();
 
+  const _isVisible = () => {
+    for (let i = 0; i < contexts.length; i++) {
+      const context = contexts[i];
+      const windowHandle = context.getWindowHandle();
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -1549,14 +1561,24 @@ const _startRenderLoop = () => {
     }
 
     // wait for next frame
-    immediate = setImmediate(_renderLoop);
+    if (_isVisible()) {
+      loop = setImmediate(_renderLoop);
+      cancelLoop = clearImmediate;
+    } else {
+      loop = setTimeout(_renderLoop, 100);
```
modulesio: This looks like a good design, but this part worries me. From my understanding, this means that there is a new `100ms` delay before we enter a VR render loop.

Also, we are adding more test processing to the render loop here that I don't think is really needed.

Can we instead perhaps start/stop the render loop when there are windows to render, rather than checking and delaying each frame?

modulesio:
Fixed with https://github.com/exokitxr/exokit/pull/1162.

PR #916, opened 2019-04-14
Various canvas fixes and enhancements

### LINKED ISSUE #907: Exokit CLI uses 100% CPU

shawwn:
This PR implements most of the remaining canvas API and fixes a few existing bugs.

### DIFF
--- a/deps/exokit-bindings/canvas/src/web_color.cpp
+++ b/deps/exokit-bindings/canvas/src/web_color.cpp
@@ -2,5 +2,5 @@
 
 canvas::web_color canvas::web_color::from_string(const char *str) {
   CSSColorParser::Color color(CSSColorParser::parse(str));
-  return web_color(color.r, color.g, color.b, (unsigned char)(color.a * 255.0));
+  return web_color(color.r, color.g, color.b, (unsigned char)(color.a * 255.0 + 0.5 - 1e-5));
 }
--- a/deps/exokit-bindings/canvascontext/include/canvas-context.h
+++ b/deps/exokit-bindings/canvascontext/include/canvas-context.h
@@ -90,6 +90,14 @@ public:
   static NAN_SETTER(LineWidthSetter);
   static NAN_GETTER(StrokeStyleGetter);
   static NAN_SETTER(StrokeStyleSetter);
+  static NAN_GETTER(ShadowColorGetter);
+  static NAN_SETTER(ShadowColorSetter);
+  static NAN_GETTER(ShadowBlurGetter);
+  static NAN_SETTER(ShadowBlurSetter);
+  static NAN_GETTER(ShadowOffsetXGetter);
+  static NAN_SETTER(ShadowOffsetXSetter);
+  static NAN_GETTER(ShadowOffsetYGetter);
+  static NAN_SETTER(ShadowOffsetYSetter);
   static NAN_GETTER(FillStyleGetter);
   static NAN_SETTER(FillStyleSetter);
   static NAN_GETTER(FontGetter);
@@ -160,12 +168,18 @@ public:
   virtual ~CanvasRenderingContext2D();
 
 // protected:
+
+  void Configure();
+
   bool live;
   NATIVEwindow *windowHandle;
   GLuint tex;
   sk_sp<GrContext> grContext;
   sk_sp<SkSurface> surface;
   SkPath path;
+  float shadowBlur;
+  float shadowOffsetX;
+  float shadowOffsetY;
   SkPaint strokePaint;
   SkPaint fillPaint;
   SkPaint clearPaint;
@@ -173,6 +187,9 @@ public:
   std::string textAlign;
   TextBaseline textBaseline;
   Direction direction;
+  Nan::Persistent<Value> jsShadowColor;
+  Nan::Persistent<Value> jsFillStyle;
+  Nan::Persistent<Value> jsStrokeStyle;
 
   /* friend class Image;
   friend class ImageData;
--- a/deps/exokit-bindings/canvascontext/include/canvas-context.h
+++ b/deps/exokit-bindings/canvascontext/include/canvas-context.h
@@ -33,18 +33,19 @@ class ImageBitmap;
 class Path2D;
 class CanvasGradient;
 
-enum class TextBaseline {
-  TOP,
-  HANGING,
-  MIDDLE,
-  ALPHABETIC,
-  IDEOGRAPHIC,
-  BOTTOM,
-};
-enum class Direction {
-  LEFT_TO_RIGHT,
-  RIGHT_TO_LEFT,
-};
+static const char* kTextBaseline_TOP = "top";
+static const char* kTextBaseline_HANGING = "hanging";
+static const char* kTextBaseline_MIDDLE = "middle";
+static const char* kTextBaseline_ALPHABETIC = "alphabetic";
+static const char* kTextBaseline_IDEOGRAPHIC = "ideographic";
+static const char* kTextBaseline_BOTTOM = "bottom";
+static const char* kTextDirection_LEFT_TO_RIGHT = "ltr";
+static const char* kTextDirection_RIGHT_TO_LEFT = "rtl";
+static const char* kTextAlign_LEFT = "left";
+static const char* kTextAlign_RIGHT = "right";
+static const char* kTextAlign_CENTER = "center";
+static const char* kTextAlign_START = "start";
+static const char* kTextAlign_END = "end";
 
 class CanvasRenderingContext2D : public ObjectWrap {
 public:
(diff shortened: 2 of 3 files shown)


deps/exokit-bindings/canvascontext/src/canvas-context.cc
```diff
@@ -1409,22 +1549,31 @@ sk_sp<SkImage> CanvasRenderingContext2D::getImage(Local<Value> arg) {
...
   tex(0),
-  lineHeight(1)
+  lineHeight(1),
+  shadowBlur(0),
+  shadowOffsetX(0),
+  shadowOffsetY(0)
 {
   // flipCanvasY(surface->getCanvas());
 
+  jsStrokeStyle.Reset(JS_STR("#000000"));
```
shawwn: FWIW, these default values are equal to Chrome's for freshly-created canvases.

deps/exokit-bindings/canvascontext/src/canvas-context.cc
```diff
@@ -361,6 +391,8 @@ NAN_METHOD(CanvasRenderingContext2D::New) {
   Nan::SetAccessor(ctxObj, JS_STR("textBaseline"), TextBaselineGetter, TextBaselineSetter);
   Nan::SetAccessor(ctxObj, JS_STR("direction"), DirectionGetter, DirectionSetter);
 
+  ctxObj->Set(JS_STR("font"), JS_STR("10px sans-serif"));
```
shawwn: This is Chrome's default font for freshly-created canvases.

deps/exokit-bindings/canvascontext/include/canvas-context.h
```diff
@@ -160,19 +168,28 @@ class CanvasRenderingContext2D : public ObjectWrap {
...
+  float shadowOffsetX;
+  float shadowOffsetY;
   SkPaint strokePaint;
   SkPaint fillPaint;
   SkPaint clearPaint;
   float lineHeight;
   std::string textAlign;
   TextBaseline textBaseline;
   Direction direction;
+  Nan::Persistent<Value> jsShadowColor;
```
modulesio: Would rather not store any JS-level things here. If necessary, we can store the e.g. string value and reconstruct the JS value again later on get.

There are implications to attaching GC lifetime of these to the lifetime of `CanvasRenderingContext2D` that we should avoid.

deps/exokit-bindings/canvas/src/web_color.cpp
```diff
@@ -1,6 +1,7 @@
 #include <web_color.h>
+#include <cfloat>
```
modulesio: Is this needed?

deps/exokit-bindings/canvascontext/src/canvas-context.cc
```diff
@@ -426,25 +460,126 @@ NAN_SETTER(CanvasRenderingContext2D::StrokeStyleSetter) {
...
 
     CanvasPattern *canvasPattern = ObjectWrap::Unwrap<CanvasPattern>(Local<Object>::Cast(value));
     context->fillPaint.setShader(canvasPattern->getShader());
+    context->jsStrokeStyle.Reset(value);
   } else {
     Nan::ThrowError("strokeStyle: invalid arguments");
   }
 }
 
+void CanvasRenderingContext2D::Configure() {
```
modulesio: I don't think this is an appropriate name for this method. It's not really configuring anything, and strictly deals with shadows.

Maybe a local function (not method) like `makeDrawLooper` would be more appropriate?

deps/exokit-bindings/canvascontext/src/canvas-context.cc
modulesio: If it works!

deps/exokit-bindings/canvas/src/web_color.cpp
shawwn: Nope! Just nuked it.

deps/exokit-bindings/canvascontext/src/canvas-context.cc
shawwn: It could be a local function, but then the member variables would need to stay public. I see that `protect:` is commented out in the header.

(I agree about the name though.)

deps/exokit-bindings/canvascontext/include/canvas-context.h
shawwn: That's fine for shadow color, but unfortunately the other values can be either a string / canvasgradient / canvaspattern.

modulesio: In that case perhaps a struct?

deps/exokit-bindings/canvascontext/src/canvas-context.cc
```diff
@@ -264,38 +310,36 @@ float getFontBaseline(const SkPaint &paint, const TextBaseline &textBaseline, fl
...
-    case TextBaseline::BOTTOM:
-    case TextBaseline::IDEOGRAPHIC:
-      return -fontMetrics.fDescent;
-    case TextBaseline::MIDDLE:
-      return -fontMetrics.fDescent + fontMetrics.fCapHeight * lineHeight / 2.0f;
-    case TextBaseline::ALPHABETIC:
-    default:
-      // Do nothing.
-      break;
+  if (textBaseline == kTextBaseline_TOP) {
```
modulesio: How was this change derived? Since it's quite different from the (possibly wrong) code before.

deps/exokit-bindings/canvascontext/src/canvas-context.cc
```diff
@@ -583,25 +745,50 @@ NAN_SETTER(CanvasRenderingContext2D::FontWeightSetter) {
...
-    double lineHeight = TO_DOUBLE(value);
-
+    double lineHeight = 1;
+    if (value->IsString()) {
+      Nan::Utf8String text(value);
+      std::string lineHeightString(*text, text.length());
+      if (lineHeightString == "normal") {
+        lineHeight = 1;
+      } else {
+        Nan::ThrowError("lineHeight: invalid arguments");
```
modulesio: Is it correct to be throwing but still setting local `lineHeight`? I would expect either or...

deps/exokit-bindings/canvascontext/include/canvas-context.h
```diff
@@ -31,18 +33,19 @@ class ImageBitmap;
...
-  MIDDLE,
-  ALPHABETIC,
-  IDEOGRAPHIC,
-  BOTTOM,
-};
-enum class Direction {
-  LEFT_TO_RIGHT,
-  RIGHT_TO_LEFT,
-};
+static const char* kTextBaseline_TOP = "top";
```
modulesio: Why are text constants used here, instead of an enum? There seems to be pointer comparisons against these, which is fine, but things could get confusing later when textBaseline comes from user input -- in which case such comparisons would not do the right thing.

If we need to print these, perhaps an enum + stringifier is the more foolproof option.

deps/exokit-bindings/canvascontext/src/canvas-context.cc
shawwn: I tried to match http://tutorials.jenkov.com/html5-canvas/text.html

image

deps/exokit-bindings/canvascontext/src/canvas-context.cc
shawwn: that's a mistake.

deps/exokit-bindings/canvascontext/include/canvas-context.h
shawwn: It avoids large if-this-then-string chains in every getter.

textBaseline already comes from user input, doesn't it? That was partly why I wanted to convert everything to strings. Then I got rid of the string comparisons and ended up with that.


The downside is a lack of type safety. And I dislike that if anyone makes a mistake, the result is a crash instead of an invalid value. I'll change it back later.

deps/exokit-bindings/canvascontext/src/canvas-context.cc
modulesio: Does this PR match up with that render?

deps/exokit-bindings/canvascontext/src/canvas-context.cc
```diff
@@ -1408,16 +1673,26 @@ sk_sp<SkImage> CanvasRenderingContext2D::getImage(Local<Value> arg) {
...
+  shadowOffsetX(0),
+  shadowOffsetY(0),
+  textBaseline(kTextBaseline_ALPHABETIC),
+  textAlign(kTextAlign_START),
+  direction(kTextDirection_LEFT_TO_RIGHT)
 {
   // flipCanvasY(surface->getCanvas());
 
   strokePaint.setTextSize(12);
+  strokePaint.setColor(0xff000000);
```
modulesio: Is red stroke the correct default to use?

deps/exokit-bindings/canvascontext/src/canvas-context.cc
```diff
@@ -1408,16 +1673,26 @@ sk_sp<SkImage> CanvasRenderingContext2D::getImage(Local<Value> arg) {
...
   strokePaint.setStyle(SkPaint::kStroke_Style);
   strokePaint.setBlendMode(SkBlendMode::kSrcOver);
 
   fillPaint.setTextSize(12);
+  fillPaint.setColor(0xff0000ff);
```
modulesio: Is this the right default fill to use?

modulesio:
Closing in favor of https://github.com/exokitxr/exokit/pull/940.

PR #873, opened 2019-03-20
Node 12 (JITless) support

modulesio:
V8 now supports [running in JITless mode](https://v8.dev/blog/jitless) and there is a PR to [bring it into node](https://github.com/nodejs/node/pull/26685).

With this feature, we can run Exokit on more platforms, including iOS, which disallow the executable writable memory policy that V8 uses by default.

The main task for us in exoland is to make sure we are compatible with Node 12. The V8 API changed a bit re: `v8::Handle` here, and some deprecated methods were removed. This PR makes the necessary changes to exokit and collects the needed external module updates.

Notably we would need to do this work anyway when Node 12 comes around in April.

Continuation of https://github.com/webmixedreality/exokit/pull/871; thanks to @shawwn for the research and bootstrapping that work.

PR #760, opened 2019-01-24
Multithreading (reality layers)

modulesio:
Exokit's architecture has up until this point been mostly singlethreaded. This means that the render loop, and all user code animation frames ran in the same thread. This is mostly because that is how `node` (used to) work.

However, this is problematic for several reasons:
- Web APIs are architected around the idea that user code runs alone in its own thread -- for example, network requests, image decode, web workers, service workers, audio service threads all expect to be run in parallel to user code, not in the user thread 
- User threads can hold up the processing of network requests, resulting in needlessly reduced network throughput when doing heavy user work such as rendering
- User thread lockups can lock up Exokit engine entirely
- User contexts are very hard to track and destroy without encapsulating them inside of a true isolated V8 context environment
- Bugs from one context can leak to other contexts
- When Exokit loads and composes multiple WebXR iframes into a single world view ("reality tabs"), we need to make sure slow apps do not hold up fast apps -- this implies a desynchronized render loop with reprojection for apps that can't keep up, and that's only possible if we run each app in its own thread separate from the top level render loop

Node 11+12 have [full threading support](https://nodejs.org/api/worker_threads.html), which mostly follows the WebWorker API.

This PR leverages `worker_threads` to run all user window code in parallel, separate from the render loop. The architecture is as follows:

- These is still a top level render loop, which has a list of top-level `window`s, corresponding to web `window` contexts
- The render loop asynchronously dispatches the XR three-phase events to the `window` contexts: `waitPose`, `tickAnimationFrame`, `submitFrame`. This is effected by posting messages down child `window`s, and capturing the response in a `Promise` which we can `.then` on
- Since each window can have child 3D `<iframe>`s which it is composing into its scene, each window is also its own render loop, and proxies the three-phase XR to its child windows. This forms a virtual render tree structure made of `window` contexts
- Use GLSync objects from the child `window` renders, waited on by its owning `window`, in order to effect perfect frame synchronization for apps in the ideal case
- If an app does not meet the frame deadline, we `setTimeout` and continue rendering without it, and do not dispatch any more `requestAnimationFrame` to it -- whenever the app finishes the render and resolves the corresponding `Promise`, we will pick up the frame on the next available render loop tick and resume further `requestAnimationFrame` to it
- Everything else works normally as if we had multiple V8/GC contexts, which is friendly to OS threading, scheduling and allocation

[...]

PR #919, opened 2019-04-15
Node 12 compatibility fixes

modulesio:
[Node 12 is coming soon](https://github.com/nodejs/node/issues/25082) and exokit already did the [bulk of the work to support the move](https://github.com/exokitxr/exokit/pull/873).

However, recent PR's like Oculus Desktop support have added new incompatibilies with Node 12. This PR fixes those.


modulesio:
Note that `vm-one` still won't build with Node 12 -- either we need to patch that source, or we need to ship [Reality Tabs](https://github.com/exokitxr/exokit/pull/760).

PR #922, opened 2019-04-15
clarifcations for MagicLeap build process

Utopiah:
few details to hopefully make the process easier for newcomers


BUILDING.md
```diff
@@ -40,21 +42,21 @@ Exokit is a node module. The core is written in Javascript, with native bindings
 
 > *NOTE: Due to combined usage of Ubuntu and Windows tools, you must check out Exokit to somewhere reachable from both; the easiest way is to do your `git clone` inside Ubuntu bash, since for example Ubuntu can reach `C:\exokit` as `/mnt/c/exokit`. (included in instructions below)*
 
-### Procedure (signed)
+### Procedure (unsigned)
```
modulesio: Thanks, just one thing: `signed` is the common case so we should probably keep that as the first one listed.

Utopiah: Fixed accordingly

modulesio:
Thanks!

PR #929, opened 2019-04-16
Oculus desktop direct rendering

modulesio:
The initial Oculus desktop binding [uses a blit phase](https://github.com/exokitxr/exokit/blob/8d8ced6c427d0fe455d24c55ed0cabe4ed9c5c13/deps/oculus/src/ovrsession.cpp#L460) to copy the rendered framebuffer into the headset textures.

It seems this is actually not necessary -- we can save the copy by rendering to the headset swapchain texture instead, with the proper layout configuration.

This should result in slightly increased throughput. However, the bigger benefit is detaching GL processing from the HMD submit phase, which allows us to have a cleaner parallelized render loop implementation.

PR #930, opened 2019-04-16
Oculus Go direct rendering

modulesio:
Sibling of Oculus desktop direct rendering https://github.com/exokitxr/exokit/pull/929.

This refactors the Oculus mobile swap chain to render directly to the headset swap chain texture instead of blitting to it from our own framebuffer.

This should save some non-trivial performance, since it saves a full synchronous memory copy each frame.

### DIFF
--- a/deps/exokit-bindings/egl/src/egl.cc
+++ b/deps/exokit-bindings/egl/src/egl.cc
@@ -247,8 +247,6 @@ NAN_METHOD(Create3D) {
     glGenTextures(sizeof(framebufferTextures)/sizeof(framebufferTextures[0]), framebufferTextures);
   }
 
-  windowsystembase::InitializeLocalGlState(gl);
-
   GLuint vao;
   glGenVertexArrays(1, &vao);
   glBindVertexArray(vao);
--- a/deps/exokit-bindings/glfw/src/glfw.cc
+++ b/deps/exokit-bindings/glfw/src/glfw.cc
@@ -1210,8 +1210,6 @@ NAN_METHOD(Create3D) {
     glfwSetCursorEnterCallback(windowHandle, cursorEnterCB);
     glfwSetScrollCallback(windowHandle, scrollCB);
     
-    windowsystembase::InitializeLocalGlState(gl);
-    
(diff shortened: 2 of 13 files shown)


modulesio:
This should now be correctly blitting directly to the Oculus Go headset texture.

modulesio:
This branch works, but I'm seeing some time-reversal in the composition. That's probably something in the swap chain iteration.

We'll need to fix this regardless for the reality tabs case, but I'll hold off on the merge for now.

modulesio:
Although this PR nominally works, it has an issue where the hardware (Oculus Go in this case) will re-use and old frame every few frames ("time travel").

This is not related to headpose, but can be seen when looking at an animated model.

My hypothesis is that a previous frame in the swap chain is being submitted before the render completes. Notably adding `glFinish` does not help.

However, the code in `master` does not have this issue, so the problem must be in this diff.

To repro, build this branch and look at the exobot in `tutorial.html`.

modulesio:
This is probably not related to this branch, but I suspect reality tabs aren't working in Oculus Go because of usage of accelerated canvas2d in Skia.

modulesio:
I found a way to make this work with a single-texture composition for Oculus Mobile.

It turns out that it _is_ possible to do a final blit-less composition to the eyes, by remultiplying the projection matrix used in the layer submission. This way, we can use a triple-buffered rotating swap chain injected into the windows, and simply set the matrices to render it at the top level submit.

If the user does not request antialiasing it is theoretically possible to do a completely blitless render, at least on Oculus.

modulesio:
Closing in favor of https://github.com/exokitxr/exokit/pull/1071.

modulesio:
Reopening because this PR has a useful GL shader optimization -- the composition shaders can be elided until they are actually used.

See: https://github.com/exokitxr/exokit/pull/930/files#diff-9072aaf78a49fe573d7263fd2ac67503R54

modulesio:
The useful code was extracted into https://github.com/exokitxr/exokit/pull/1146, so closing this again.

PR #931, opened 2019-04-16
Consider multiple calls to requestPresent with different canvases

dmarcos:
A-Frame master attaches a `vrdisplayactivate` listener early, before the scene loads. If the event triggers, a dummy canvas is used to call `requestPresent` immediately to fit within the time limits that some browsers have to exercise the option of invoking `requestPresent` without user gesture requirement. requestPresent is called again when the scene loads and the real canvas and associated webgl context are initialized. Current code in master will ignore the second call because a glContext is already defined resulting in the scene not presenting.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -480,7 +480,7 @@ if (nativeBindings.nativeOculusVR) {
     const layer = layers.find(layer => layer && layer.source && layer.source.tagName === 'CANVAS');
     if (layer) {
       const canvas = layer.source;
-      if (!vrPresentState.glContext) {
+      if (vrPresentState.glContext !== canvas._context) {
         let context = canvas._context;
         if (!(context && context.constructor && context.constructor.name === 'WebGLRenderingContext')) {
           context = canvas.getContext('webgl');
--- a/src/index.js
+++ b/src/index.js
@@ -598,7 +598,7 @@ if (nativeBindings.nativeOpenVR) {
     if (layer) {
       const canvas = layer.source;
 
-      if (!vrPresentState.glContext) {
+      if (vrPresentState.glContext !== canvas._context) {
         let context = canvas._context;
         if (!(context && context.constructor && context.constructor.name === 'WebGLRenderingContext')) {
           context = canvas.getContext('webgl');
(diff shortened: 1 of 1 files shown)


modulesio:
Thanks, I can see now why https://github.com/exokitxr/exokit/pull/920 made some sense for this case -- it would always trigger in the oculus case.

src/index.js
```diff
@@ -480,7 +480,7 @@ if (nativeBindings.nativeOculusVR) {
     const layer = layers.find(layer => layer && layer.source && layer.source.tagName === 'CANVAS');
     if (layer) {
       const canvas = layer.source;
-      if (!vrPresentState.glContext) {
+      if (vrPresentState.glContext !== canvas._context) {
```
modulesio: Presuming this is correct, we'll probably need a similar condition for the other platforms:

- https://github.com/exokitxr/exokit/blob/e372292cf1caaa436862d2f8b3768a6804d9fa68/src/index.js#L601
- https://github.com/exokitxr/exokit/blob/e372292cf1caaa436862d2f8b3768a6804d9fa68/src/index.js#L768

As well as upcoming Oculus Go/Mobile.

modulesio: Looking at the below code I think this is the correct approach.

However, doing this naively will leak the previous presenting session and associated resources, such as the swap chain. We should take care to tear that down, or reuse it, or else there can be unpredictible effects.

The most correct way would be to tar down the session and re-make it, but [that is not implemented for the Oculus case](https://github.com/exokitxr/exokit/pull/931/files#diff-1fdf421c05c1140f6d71444ea2b27638L592).

Additionally, is is the intent to [clean up the way swap chains work](https://github.com/exokitxr/exokit/pull/929) which would affect this code path.

So maybe if we get this check implemented for all devices in this PR, we can open an issue for the resource cleanup?

modulesio: https://github.com/exokitxr/exokit/issues/932

dmarcos:
I added the old glContext checks to the OpenVR and ML branches. I cannot test in ML.

src/index.js
```diff
@@ -765,7 +765,7 @@ if (nativeBindings.nativeMl) {
     if (layer) {
       const canvas = layer.source;
 
-      if (!mlPresentState.mlGlContext) {
+      if (mlPresentState.glContext !== canvas._context) {
```
modulesio: I think it's called `mlGlContext` in the code.

dmarcos: Yeah sorry. Just renamed. Maybe good to eventually rename all of them? Anything special with mlGLContext?

modulesio: Nothing special, it's a legacy artifact of the code. It just won't function unless the names match.

Issue #935, opened 2019-04-17
Magic Leap direct rendering


modulesio:
As we are doing with [Oculus Rift](https://github.com/exokitxr/exokit/pull/929) and [Oculus Go](https://github.com/exokitxr/exokit/pull/930), we should make the Magic Leap fbo rendering case work with a single blit to the headset textures.

This shouldn't be too bad since the Magic Leap API already uses a texture array, which can be bound to two fbos as in the Oculus Go case.

Blocker for the performance optimization in https://github.com/exokitxr/exokit/pull/760.

modulesio:
Closing as outdated.

PR #938, opened 2019-04-17
Magic Leap direct headset blit

modulesio:
Sibling of https://github.com/exokitxr/exokit/pull/929, https://github.com/exokitxr/exokit/pull/930.

Fixes https://github.com/exokitxr/exokit/issues/935.


modulesio:
For oculus go we have a cleaner blitting strategy where the swap chain is passed and flipped on the compositor window side.

Well probably want to rewrite this pr in terms of that implementation.

modulesio:
This is related to https://github.com/exokitxr/exokit/pull/1122, since the Magic Leap SDK wants a multiview-like texture array.

However, there is nothing interesting in this PR as is. When we ship multiview we can rethink a direct headset blit for ML.

Issue #934, opened 2019-04-17
Meshing user API figure-outage


modulesio:
Thus far we have handled the GL buffers for the meshing system at the engine level. This has the benefit of being highly optimized, and allows for doing things like occlusion at the frame level without the user code doing any work.

However, this assumes a lot about how the user wants to use the mesh data. In particular, it is wasteful if the user wants to do their own buffer processing by loading the Javascript Float32Array into WebGL themselves. This also adds significant complexity and overhead in the reality tabs case, since we potentially need to submit the data to multiple GL contexts, which is both slow and error-prone.

I'm leaning to making the API such that it provides the mesh as a `Float32Array` to the user on request, without doing any GL processing ourselves.

This goes beyond our Magic Leap API binding, since there are other SDK's that present mesh data, so we should standardize on a simple meshing mechanism for all platforms.

[This incubation repository](https://github.com/immersive-web/real-world-geometry) in `immersive-web` is relevant spec-wise.

Figuring this out is a blocker for https://github.com/exokitxr/exokit/pull/760.

modulesio:
Partial implementation of this idea: https://github.com/exokitxr/exokit/pull/764/files

modulesio:
The WebXR extension API experiments for this are taking shape in https://github.com/exokitxr/exokit/pull/1002.

Since there is a centralized place for that implementation now, we probably don't need this issue. The Immersive Web links remain relevant.

PR #939, opened 2019-04-17
[ML] Mesh buffers de-gl

modulesio:
Removes the dependency of the Magic Leap binding on doing GL processing in the render loop.

The magic leap bindings previously dabbled in setting the GL buffers for features such as meshing, for performance reasons.

However, with user window parallelization, whose main point is to reduce GL context switching outside of user code, this makes a lot less sense. Additionally, the previous API led to user confusion re: buffer validity, and for at least one user this was a performance complaint.

Therefore, this PR drops this feature in favor of letting the user set the GL buffers themselves, if they so wish.

Fixes https://github.com/exokitxr/exokit/issues/934.
Supercedes https://github.com/exokitxr/exokit/pull/764.

### DIFF
--- a/deps/exokit-bindings/magicleap/include/magicleap.h
+++ b/deps/exokit-bindings/magicleap/include/magicleap.h
@@ -133,15 +133,14 @@ public:
 class MeshBuffer {
 public:
   MeshBuffer();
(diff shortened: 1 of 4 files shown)


modulesio:
This branch can land into `user-parallel` since that is the main consumer.

PR #944, opened 2019-04-19
Add Oculus Mobile APK build to Travis CI

chrisplatorres:
This PR adds the dockerfile and update travis config to add the Oculus Mobile APK build to CI.


android/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM debian:latest
+
+RUN apt-get update && \
+  apt-get install -y \
```
modulesio: Do we know if these are actually needed/can we whittle it down to the minimun? Some of these seem like they wouldn't be used and we can save CI time by not including them.

chrisplatorres: Good point, i'll check

android.Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM debian:latest
+
+RUN apt-get update && \
```
rzr: apt-get update -y 
will be safer here

android/Dockerfile
chrisplatorres: 9f5c7d8

chrisplatorres:
Should now be ready to test on Travis CI.

modulesio:
Note that this seems broken on Travis for technical reasons:

```
[=============================          ] 74% Unzipping... android-ndk-r19c/tool
[=============================          ] 74% Unzipping... android-ndk-r19c/tool
[=============================          ] 74% Unzipping... android-ndk-r19c/tool
[=============================          ] 74% Unzipping... android-ndk-r19c/tool
[=============================          ] 74% Unzipping... android-ndk-r19c/tool
[=============================          ] 74% Unzipping... android

The job exceeded the maximum log length, and has been terminated.
```

I think we need to silence some of this output.

chrisplatorres:
This is a Travis CI [common build error](https://docs.travis-ci.com/user/common-build-problems/#log-length-exceeded) 
> The log for each build is limited to approximately 4 Megabytes. When it reaches that length the build is terminated and you’ll see the following message at the end of your build log:

Going to silence some of the heavy sdkmanager logging. If there isn't a reasonable amount of silencing that fixes this, we can output to a log and keep Travis awake by periodically logging:

```
  - docker build -t exokit . > build.log 2>&1
  - while sleep 5m; do echo "=====[ $SECONDS seconds, exokit still building... ]====="; done &
  # Killing background sleep loop
  - kill %1
```

Issue #951, opened 2019-04-20
Failed to load Cryptovoxels


madjin:
**Describe the bug**
 What happened, and what did you expect to happen?
 Run `node --experimental-worker . https://cryptovoxels.com/play` and get an error that: `ReferenceError: HTMLBodyElement is not defined`
 
**To Reproduce**
Steps or commands ran to reproduce the behavior:
1. Go to 'terminal (MINGW64)'
2. Type `node --experimental-worker . https://cryptovoxels.com/play`
3. See error

**Additional context**
If you walk or fly around the city the url changes, might be of use to know this to test urls related to cryptovoxels.

**System information:**
 - OS: Windows 10
 
 - Exokit Version: d38467d
 
 - Node version: v11.6.0
 
 - Npm version/build log if applicable
 
 - Is your Exokit downloaded or built?
Built from source

**Screenshots**
output log: https://pastebin.com/zkEHWMS7

modulesio:
Cool, this one sounds relatively easy. [We even have `HTMLBodyElement`](https://github.com/exokitxr/exokit/blob/master/src/Window.js#L112), it just wasn't on `window`.

I don't think https://developer.mozilla.org/en-US/docs/Web/API/HTMLBodyElement has much, if any necessary functionality to it.

modulesio:
Worth trying this again, since https://github.com/exokitxr/exokit/pull/952 was merged.

madjin:
Getting a new error `ReferenceError: HTMLDivElement is not defined`

output log: https://pastebin.com/wF7yGaG5

modulesio:
With https://github.com/exokitxr/exokit/pull/953, we get further into widgets loading:

```
Error: #discord-widget element does not exist!
    at Object.<anonymous> (/assets/e5d3f5da2fab98ab650a.js:30:23854)
    at n (/assets/e5d3f5da2fab98ab650a.js:1:110)
    at /assets/e5d3f5da2fab98ab650a.js:1:910
    at /assets/e5d3f5da2fab98ab650a.js:1:920
    at Object._runJavascript (C:\Users\avaer\Documents\GitHub\exokit5\src\utils.js:71:15)
    at ownerDocument.defaultView.fetch.then.then.s (C:\Users\avaer\Documents\GitHub\exokit5\src\DOM.js:1675:17)
    at processTicksAndRejections (internal/process/next_tick.js:81:5)
TypeError: Cannot read property 'responseStart' of undefined
[...]

PR #953, opened 2019-04-20
Add missing HTML DOM elements

modulesio:
Partial fix for https://github.com/exokitxr/exokit/issues/951.


modulesio:
Need to investigate the [`id`-related test failure](https://api.travis-ci.org/v3/job/522362335/log.txt) before merging. It might have to do with some of the subsequent throws in https://github.com/exokitxr/exokit/issues/951.

PR #965, opened 2019-04-22
Remove ndb for debugging

chrisplatorres:
This PR removes ndb from package.json and README. I think Exokit debugging has generally been done with gdb/winDbg. A doc page dedicated to debugging might be necessary in this case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,24 +160,6 @@ npm install
 node --experimental-worker . <url> # or node --experimental-worker . -h for home
 ```
 
-## Debugging
-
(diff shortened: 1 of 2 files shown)


modulesio:
The other side to this is site debugging (attaching with DevTools), not necessarily debugging the exokit engine core.

modulesio:
I’m thinking it might be worth integrating ndb flow into the launcher (https://github.com/exokitxr/exokit/pull/974) rather than dropping it.

chrisplatorres:
I agree. Going to close this for now.

PR #969, opened 2019-04-23
Initial support for Oculus Quest controllers (6DOF tracking)

dmarcos:
This PR focus on having two different paths for 3DOF controllers (GearVR, Oculus GO) and 6DOF (Quest). I'll follow up with PRs to fix the buttons mapping and different ids for each controller so applications can discriminate and use the appropriate model.

### DIFF
--- a/deps/oculus-mobile/src/oculus-context.cpp
+++ b/deps/oculus-mobile/src/oculus-context.cpp
@@ -83,7 +83,7 @@ void OculusMobileContext::handleAppCmd(struct android_app *app, int32_t cmd) {
 	}
 }
 
-Local<Function> OculusMobileContext::Initialize() {  
+Local<Function> OculusMobileContext::Initialize() {
   Nan::EscapableHandleScope scope;
 
   // constructor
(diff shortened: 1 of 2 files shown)


deps/oculus-mobile/src/oculus-context.cpp
```diff
@@ -335,9 +340,13 @@ NAN_METHOD(OculusMobileContext::WaitGetPoses) {
...
                 ovrTracking tracking;
                 vrapi_GetInputTrackingState(oculusMobileContext->ovrState, capsHeader.DeviceID, predictedDisplayTime, &tracking);
-                ovrVector3f pivot{0.2, 1, 0};
-                ovrVector3f point{0.2, 1, -0.2};
-                tracking.HeadPose.Pose.Position = ovrVector3f_RotateAboutPivot(&tracking.HeadPose.Pose.Orientation, &pivot, &point);
+                // 3DOF Controller Tracking
+                if (remoteCaps.ControllerCapabilities & ovrControllerCaps_ModelOculusGo ||
+                    remoteCaps.ControllerCapabilities & ovrControllerCaps_ModelGearVR) {
+                  ovrVector3f pivot{-0.2, 1, 0};
+                  ovrVector3f point{-0.2, 1, -0.2};
```
modulesio: This one is for the right hand so it should be pivoted `0.2, 1, -0.2` I think.

deps/oculus-mobile/src/oculus-context.cpp
```diff
@@ -335,9 +340,13 @@ NAN_METHOD(OculusMobileContext::WaitGetPoses) {
               } else if (remoteCaps.ControllerCapabilities & (ovrControllerCaps_RightHand|ovrControllerCaps_ModelOculusGo)) {
                 ovrTracking tracking;
                 vrapi_GetInputTrackingState(oculusMobileContext->ovrState, capsHeader.DeviceID, predictedDisplayTime, &tracking);
-                ovrVector3f pivot{0.2, 1, 0};
-                ovrVector3f point{0.2, 1, -0.2};
-                tracking.HeadPose.Pose.Position = ovrVector3f_RotateAboutPivot(&tracking.HeadPose.Pose.Orientation, &pivot, &point);
+                // 3DOF Controller Tracking
+                if (remoteCaps.ControllerCapabilities & ovrControllerCaps_ModelOculusGo ||
```
modulesio: This condition looks ok, but [the one above](https://github.com/exokitxr/exokit/pull/969/files#diff-897f9a3834c7903b1e52128799c158b0R340) I think would preclude GearVR support. So if we should probably add that check above as well.

dmarcos: I removed the condition, is it really need it?

deps/oculus-mobile/src/oculus-context.cpp
dmarcos: fixed

deps/oculus-mobile/src/oculus-context.cpp
modulesio: No it wouldn't be needed, but as far as I remember it won't work on Oculus Go otherwise -- the controller is not actually associated with a hand.

dmarcos: I believe they do but not 100% sure. I would have to try. You can change handednnes in the Oculus app on your phone for the Go.

modulesio: Could you give it a shot? I think that was the only reason for the original code -- the remote counted as neither hand.

dmarcos: I removed all 3DOF specific to use the straight pose that comes from the SDK. I tested in both Oculus Go and Quest and works fine. Go provides a hand that comes from user configuration in the Oculus smartphone app.

modulesio: Do we still get the controller elbow-modeled this way on the Go? In my testing the controller was rotating around the origin rather than in a reasonable hand position offset from the head.

dmarcos: Based on my tests it tracks the same way as the one in the Oculus menu. The hand switches from left to right when changing settings. It should not make any difference but I'm testing with A-Frame master and the [laser-control example](https://github.com/aframevr/aframe/tree/master/examples/test/laser-controls)

modulesio: Great, thanks for testing.

Issue #886, opened 2019-04-03
Testing suite to maintain exokit-supported list of links


chrisplatorres:
This issue to to add a testing suite that tests [reality tabs](https://github.com/webmixedreality/exokit/blob/ee72c113878c42a9e156665972fce00f99c84e5f/examples/realitytabs.html#L15-L83)/["app list"](https://exokit.org/docs/exokitApps/) links alongside the current CI. The ["app list"](https://exokit.org/docs/exokitApps/) links will be updated and added to reality tabs links section. The testing suite would make this list easier to maintain. 

Potential issues:
- getting a GPU CI

modulesio:
Xvfb should work as well, but it’s always finnicky. I think that’s how it currently runs.

PR #974, opened 2019-04-24
Exokit launcher v0.1

modulesio:
This adds a Chromium Embedded (powered) launcher interface to Exokit engine, for URL entry, viewing examples, and configuring engine start settings.

The interface is just a `launcher.html` site with a 2D iframe that uses `postMessage` to communicate. We could bring this into XR as well at some point.

[WIP]


chrisplatorres:
Will this launcher be desktop only? If so, will it be the same launcher across windows, macos, and linux?

modulesio:
Desktop-only for now, but maybe it will find its way into the mobile UIs such as `realitytabs.html`.

It would indeed be the same launcher on all platforms.

modulesio:
Something like this:

Screen Shot 2019-04-24 at 8 46 16 PM

chrisplatorres:
This might be more relevant for a future launcher PR, but should users be able to set multiple reality tabs to load into? Or maybe have a profile of saved reality tabs that can be loaded into, it pulls from ideas that were happening in Discord:
  
> I thought the basic UI was html, so it wouldn't need to be a single "browser", it could just load a expandable repository of whitelisted browsers hosted on github

modulesio:
Those are interesting ideas! They're interesting enough that I'd like to decouple them from the Exokit engine. The focus of the engine itself should be to run sites -- in the form of `.html` pages -- and delivering the base functionality to implement:

- A browser, both 2D and 3D
- Reality tabs UI
- World storage
- Anchor persistence
- Multiplayer
- Inventory
- Social features

However, I think we should keep Exokit's core focus as an engine and not expand its scope.

If we end up doing any (or all) of these things, they should at the minimum be constrained to  `examples/` `.html` files. More likely they would be entirely different repositories, with a different `exo`name in the constellation.

The relationship to Exokit engine would simply be that Exokit engine can run those sites. And, maybe Exokit's launcher provides a list of socialized links to try from the launcher.

chrisplatorres:
That makes a lot of sense. In that context, the `examples/` serve a very specific purpose of giving template code for different use cases.

The `examples/launcher.html` is then a template that can be forked/adapted.

> maybe Exokit's launcher provides a list of socialized links to try from the launcher.

Would this be in the form of a registry?

modulesio:
`launcher.html` can of course be forked. It's loosely coupled to the engine. It's just a bit of a special case since Exokit needs to ship with some kind of non-XR UI that's one step up from the CLI.

Re: registries, my thinking is that should be totally removed from this repository. _Maybe_ `launcher.html` would dynamically fetch it, but it's not even clear if it's desirable for Exokit to take on the task of maintaining WebXR site registry.

chrisplatorres:
As of now [this page](https://exokit.org/showcase/) in placeholder mode, but could serve as a curated list of tested / supported on Exokit engine sites. 

This of course brings up the issues with how that is maintained, if there will need to be an automatic testing suite to check the sites....

chrisplatorres:
from #965 @modulesio 
> I’m thinking it might be worth integrating ndb flow into the launcher (#974) rather than dropping it.

How should ndb be integrated here?

modulesio:
It could be an flag to run the target site in debug mode, or perhaps to "open devtools" for the current site with the devtools shortcut (Ctrl-Shift-i).

chrisplatorres:
Will the launcher retire the "Exokit Engine CLI"?

modulesio:
No, there are reasons for the CLI to exist regardless and it's in use. It might even make sense to keep it as the default thing that happens when `node` is given no exokit arguments.

Maybe the launcher itself should even expose the CLI.

However, the default entrypoint for all installers should be the launcher.

NoahSchiffman:
That's a good question! What does the launcher include so far? Looks like it can easily replace a way to use the CLI with a GUI in a more user-friendly fashion. Perhaps, we should still think about keeping the CLI though for some of the more tech savvy people that want it and like to experiment? Or is there nothing you can do on the  Launcher GUI that won't be possible on the CLI, and vice versa?

modulesio:
The intent moving forward is to make the launcher more and more friendly, and probably replace CLI for most people, though it would fundamentally just run code that could be run in the CLI anyway.

For now it's a setter for flags, but in future it could list examples, have an updater, or expose more devtools-like functionality like starting a site in `ndb`.

NoahSchiffman:
So in the future, we could be running `Reality Tabs` from the Launcher until there is a way to launch them on any headset (and also from those headsets, not from a PC) or something like that?

modulesio:
It's already possible to run the reality tabs stack from CLI, as well as WebVR/WebXR emulation with mouse and keyboard, but yes -- it's the intent to put that front and center in the launcher when we can figure out the UI for it.

That would most likely be another PR after this one.

modulesio:
For this branch, we'll need to add `create-react-app` building of [launcher/package.json](launcher/package.json) to the `npm install` build process, or else the launcher won't function.

It's either that, or we need to build manually, or set up a git hook for it. I'm not sure what the best tradeoff is.

ChrisEddy:
could push a react build to repo for people to use, and can sidestep the `npm i` process.

chrisplatorres:
> Re: registries, my thinking is that should be totally removed from this repository. Maybe launcher.html would dynamically fetch it, but it's not even clear if it's desirable for Exokit to take on the task of maintaining WebXR site registry.

In that case, #886 should be closed

Issue #972, opened 2019-04-24
MacOS CEF integration SEGV


modulesio:
On a Macbook Pro 2018 15":

```
THREE.WebGLRenderer 96dev
0   segfault-handler.node               0x00000001072950c4 _ZL16segfault_handleriP9__siginfoPv + 100
1   libsystem_platform.dylib            0x00007fff76e75b5d _sigtramp + 29
2   ???                                 0x0000000000000000 0x0 + 0
3   exokit.node                         0x000000010880469f _ZN13CefStringBaseI20CefStringTraitsUTF16E10FromStringERKNSt3__112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEE + 143
4   exokit.node                         0x0000000108804456 _ZN7browser18initializeEmbeddedERKNSt3__112basic_stringIcNS0_11char_traitsIcEENS0_9allocatorIcEEEE + 160
5   exokit.node                         0x000000010880d8e2 _ZNSt3__114__thread_proxyINS_5tupleIJNS_10unique_ptrINS_15__thread_structENS_14default_deleteIS3_EEEEZN7browser7Browser3NewERKN3Nan20FunctionCallbackInfoIN2v85ValueEEEE3$_0EEEEEPvSI_ + 73
6   libsystem_pthread.dylib             0x00007fff76e7e2eb _pthread_body + 126
7   libsystem_pthread.dylib             0x00007fff76e81249 _pthread_start + 66
8   libsystem_pthread.dylib             0x00007fff76e7d40d thread_start + 13

libc++abi.dylib: terminating
Abort trap: 6
```

Looks like the CEF build is not compatible with the `npm install` build.

modulesio:
The bug here was a missing `cef_load_library` call to load the Chromium Embbedded MacOS framework at `node_modules/native-browser-deps-macos/lib3/macos/Chromium Embedded Framework.framework/Chromium Embedded Framework`.

modulesio:
Still have this remaining error:
```
[0424/192456.963047:ERROR:icu_util.cc(136)] icudtl.dat not found in bundle
[0424/192456.963526:ERROR:icu_util.cc(172)] Invalid file descriptor to ICU data received.
```

PR #976, opened 2019-04-25
Bugfix MacOS Chromium Embedded iframes

modulesio:
Fixes https://github.com/exokitxr/exokit/issues/972.


modulesio:
For testing we can use:

```
node --experimental-worker . examples/browser.html
```

PR #978, opened 2019-04-25
Launcher HTML UI

modulesio:


launcher/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "launcher",
   "version": "0.1.0",
   "private": true,
+  "homepage": "launcher/build/",
```
modulesio: I think this would work correctly with `"homepage": "."`

Issue #982, opened 2019-04-25
Retina scaling on MacOS


modulesio:
Currently when booting Exokit on a retina display, the framebuffer size is the physical size, while we the window size is the (smaller) virtual size you asked for. We even detect this case and set the `devicePixelRatio` here: https://github.com/exokitxr/exokit/blob/0a070e6ce2b80b271bca09998e186b0a76f17f98/src/index.js#L183

This is very error-prone however, and if you forget to use `devicePixelRatio` you end up with nasty dirty framebuffer swapping (flash of unrendered content).

I think for something like exokit we should probably put this into the hands of the developer to set the size they want, so by default there is no glitching due to framebuffer-window size mismatch.

The solution might be to use `GLFW_COCOA_RETINA_FRAMEBUFFER` = `0` here: https://github.com/exokitxr/exokit/blob/0a070e6ce2b80b271bca09998e186b0a76f17f98/deps/exokit-bindings/glfw/src/glfw.cc#L1143

This would need to be tested on a proper Apple Retina display.

PR #983, opened 2019-04-25
Real devicePixelRatio

modulesio:
For a long time we've faked the `window. devicePixelRatio` property. This PR actually queries for it, while allowing Retina (scaling) mode on MacOS to function normally, by adding another framebuffer stretch blit phase to the 2D rendering case.

This allows us to get antialiasing to the main framebuffer, which improves the quality of rendering in the 2D (non-HMD) case, including WebXR emulation.

Fixes https://github.com/exokitxr/exokit/issues/982.


modulesio:
This could use a Retina display test to check whether windows are scaled properly.

PR #984, opened 2019-04-26
Return appropriate gamepad id depending on headset vendor

dmarcos:
I made Rift and Quest ids the same since there's not available model at the moment for libraries to integrate accordingly and button mapping is identical. We will have to differentiate in the future.

### DIFF
--- a/src/VR.js
+++ b/src/VR.js
@@ -634,7 +634,7 @@ let gamepads = null;
 function getGamepads(window) {
   const {oculusVRDisplay, openVRDisplay, oculusMobileVrDisplay, magicLeapARDisplay} = window[symbols.mrDisplaysSymbol];
   const presentingDisplay =
-    GlobalContext.fakeVrDisplayEnabled ||
+    (GlobalContext.fakeVrDisplayEnabled && GlobalContext.fakePresentState.fakeVrDisplay) ||
     (oculusVRDisplay.isPresenting && oculusVRDisplay) ||
     (openVRDisplay.isPresenting && openVRDisplay) ||
     (oculusMobileVrDisplay.isPresenting && oculusMobileVrDisplay) ||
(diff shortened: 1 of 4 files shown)


src/VR.js
```diff
@@ -610,37 +610,54 @@ class FakeVRDisplay extends VRDisplay {
...
+  }
+  return controllerIDs[vrDisplay.id + hand.charAt(0).toUpperCase() + hand.slice(1)];
+}
+
 let gamepads = null;
 function getGamepads(window) {
   const {oculusVRDisplay, openVRDisplay, oculusMobileVrDisplay, magicLeapARDisplay} = window[symbols.mrDisplaysSymbol];
-  if (
+  const presentingDisplay =
     GlobalContext.fakeVrDisplayEnabled ||
```
modulesio: It wouldn't be the `presentingDisplay` in the fake vr display case based on this line. We should probably point to the actual fake VR display in the case it's detected.

dmarcos: Done. Not sure if correct. not sure how to test that code path.

src/VR.js
```diff
@@ -610,37 +610,54 @@ class FakeVRDisplay extends VRDisplay {
...
+  OculusQuestLeft: 'Oculus Touch (Left)',
+  OculusQuestRight: 'Oculus Touch (Right)',
+  OculusGo: 'Oculus Go',
+  GearVR: 'Gear VR',
+  OpenVR: 'OpenVR Gamepad',
+  OpenVRTrackerID: 'Tracker'
 };
 
+function getControllerID(vrDisplay, hand) {
+  if (!vrDisplay) { return; }
```
modulesio: We always have a `vrDisplay` when requesting a controller ID so there's probably no reason to check this, right?

src/VR.js
modulesio: For posterity, it would be running `examples/realitytabs.html`, in WebXR mode.

src/VR.js
dmarcos: done

modulesio:
Thanks!

PR #986, opened 2019-04-27
Delete native app glue and utf8 bindings directories

chrisplatorres:
This PR proposes to delete native app glue and utf8 bindings directories, which are not currently being used.


chrisplatorres:
It looks like oculus mobile bindings is using the native app glue header: https://github.com/exokitxr/exokit/blob/master/deps/oculus-mobile/include/oculus-mobile.h#L10

going to revert the native app glue dir deletion

modulesio:
Can we move `android_native_app_glue.{h,c}` under `oculus-mobile`?

chrisplatorres:
The native_app_glue header and src file here is actually overriding the ANDROID_NDK one that is in the CMakeList for oculus-mobile ( https://github.com/exokitxr/exokit/blob/master/android/CMakeLists.txt#L10 )

PR #989, opened 2019-04-28
 Move root cpp files to cpp directory

chrisplatorres:
Continuation of #889 

This PR is to move the root .cpp files to a new `cpp` directory and update associated paths.


chrisplatorres:
The worth of PRs like this to "cleanup" is questionable because there ends up being a long tail of breakages with the paths.

modulesio:
Yeah, I'm not sure about these two files for now.

We can revisit this later.

PR #992, opened 2019-04-30
Export Range constructor on window

dmarcos:
Likely a regression


modulesio:
Thanks, there's probably other cases where the objects that exist but aren't exposed.

It might be worth going through the `DOM.js`/`Window.js`/`Document.js` and making sure they are all on the correct export object.

PR #994, opened 2019-04-30
DOM-to-texture devicePixelRatio

modulesio:
This adds `iframe.devicePixelRatio` support for 2D Chromium Embedded layers, making the `launcher.html` render hopefully pixel-perfect to Chrome.

This isn't any sort of final UI but it should get the point across.

Capture 20

### DIFF
--- a/deps/exokit-bindings/browser/include/browser-common.h
+++ b/deps/exokit-bindings/browser/include/browser-common.h
@@ -65,6 +65,7 @@ EmbeddedBrowser createEmbedded(
   GLuint tex,
   int width,
   int height,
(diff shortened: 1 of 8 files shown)


chrisplatorres:
Noting the missing top left icon on the Exokit window. 🤔

modulesio:
It can be set with [`glfwSetWindowIcon`](https://www.glfw.org/docs/latest/group__window.html#gadd7ccd39fe7a7d1f0904666ae5932dc5).

For that we would need to get a 48x48 non-premultiplied RGBA binary version of the exokit icon.

PR #996, opened 2019-04-30
Remove source releases from CI configs

chrisplatorres:
This PR proposes moving towards installer for platforms instead of confusing people with source releases. The releases should have per-platform installers / executables, If someone wants source there is a better path to take.


chrisplatorres:
Yes, I'll do the same for mac and linux here.

PR #1000, opened 2019-05-01
Zed SDK support

modulesio:
For passthrough AR-in-VR, this PR adds initial ZED stereo camera/meshing support on Windows.

[WIP]


chrisplatorres:
Does the Zed license allow for distribution of the lib/header files?

chrisplatorres:
> a.     The Software is licensed to You, not sold. You are licensed to use the Software only as pre-installed in Your ZED USB flash drive, and updated by STEREOLABS from time to time. You may not copy or reverse engineer the Software.

It is hard to tell based on that what is meant by "copy".

modulesio:
Interesting find. There’s ways we can deal with it if it’s a problem, like making every user do the install.

I think what this is trying to cover is selling the  zed sdk or hacking into their code — we’re not doing that. Linking to the sdk is the primary expected use of the sdk so I don’t think that can be banned.

modulesio:
This probably makes the most sense to base off of `studio`, where the meshing APIs are fleshed out.

PR #1002, opened 2019-05-01
Exokit Studio

modulesio:
Capture 23

To build the UI:

```
cd examples/launcher
npm i
npm run build
```

Then either load `examples/launcher/build/index.html` to work on the pure HTML UI or start the studio example that composites everything (`examples/studio.html`).

[WIP] based off of the reality layers branch


modulesio:
We will want package export functionality implemented from the menut; something like

```
Export > Android APK...
```

And we probably want that for all platforms.

In general we want to avoid a full build in this export process, so the easiest path is to download the latest built asset (or the one most closely matching the currently running release), and repack that binary into a new package with the user's desired app assets.

For APK's and MPK's, the path is clear: these files are just `.zip` files which are signed, so we can repack them with the app assets under `/app` and re-pack and re-sign with the SDK signing tools.

However, for desktop, which uses installers, we probably will need a way to extract the installer/DMG to get at the files (or have a separate release that contains the files in a usable form that is not an installer, like a `.zip`). We can then ship Exokit with just the needed scripts to do the packaging phase of the install (either the InnoDB install maker, or the MacOS DMG maker). And then the desktop cases reduces to the mobile one, where we are just repacking assets.

The main purpose of doing it this way is to not have any build dependencies on the user side for doing the export.

In either case we will need a way to input the user's certificate to do the signing procedure.

chrisplatorres:
We would need to make it clear that when exporting, they are pulling from latest releases, rather than using any changes that the user has made to their local Exokit Engine.

modulesio:
I don't expect the intersection of engine hackers and users of the export is very large.

If you are hacking your own version of the engine, you are probably doing so because you want to ship in your own style.

Also, I think the studio will probably end up distanced from the engine core (separate repo) so these two workflows would interact even less.

modulesio:
The studio has most of the features we wanted now:

- Fake VR device emulation
- Reality layers (3d)
- Reality layers (2d)
- Transform reality layers
- Inspect scene like a game engine
- Edit DOM for reality layers
- Add templates to scene
- Add examples to scene
- Enter fake XR
- Enter real XR

Needs some more per-headset testing, and we're blocked on `user-parallel` going in before us. 👍

chrisplatorres:
Some UI/UX things I noticed:


- right clicking shows "Back Forward Print View page source"
rightclickCapture
- ability to delete reality tabs (there is selecting and pressing delete button, maybe add right click abilities as well)
- missing resizing sidebars / console
- opening dropdown menus makes virtual display black
layCapture

- I wonder if there is a way to parse and display the html that is more readable, maybe only showing type and src name and then hiding other values in the "inspector" menu (again thinking from Unity UI perspective)
- virtual display only shows after resizing window
  
**before resize:**
beforeresCapture

**after resize:**
afteresCapture
- tabs labeling console / inspector?
tabsCapture

examples/launcher/src/components/Dom.jsx
```diff
@@ -0,0 +1,312 @@
...
+      childNodes: [
+        {
+          nodeType: 1,
+          tagName: 'BODY',
+          value: '',
+          attrs: [{
+            name: 'lol',
+            value: 'zol',
+            name: 'woot',
+            value: 'toot',
```
chrisplatorres: Are these placeholder/debugging values?

modulesio: Yes sir.

chrisplatorres:
[url bar-> reload opens a new window](https://github.com/exokitxr/exokit/pull/1002/files?file-filters%5B%5D=.css&file-filters%5B%5D=.eot&file-filters%5B%5D=.html&file-filters%5B%5D=.js&file-filters%5B%5D=.jsx&file-filters%5B%5D=.svg&file-filters%5B%5D=.ttf&file-filters%5B%5D=.woff&file-filters%5B%5D=.woff2#diff-818c6db2ba21b9def5bafb0e208c1111R192)

Looks like window.location.reload() will work now from https://github.com/exokitxr/exokit/pull/1048

- first window updates, a-painter is flickering / only flickers when mouse is in viewport mousing over the a-painter transform
- 2nd window opens and is completely black

also noticing that the right click box will lock user out of the UI until one of the options are selected

newwindCapture


------------------------------
  
In the first menu:

- New = another studio window?

------------------------------

With the transform :
- click and hold transform while moving mouse outside of window, let go of click, moving cursor in viewport still is in transform mode.

https://github.com/exokitxr/exokit/blob/9bfde3e1747b987bf08fdddb6eea7d351d9e3ac3/examples/studio.html#L1191

chrisplatorres:
>  re-sign with the SDK signing tools.
> The main purpose of doing it this way is to not have any build dependencies on the user side for doing the export.
> In either case we will need a way to input the user's certificate to do the signing procedure.

Does the SDK signing tools not count as build dependencies here?

modulesio:
> Does the SDK signing tools not count as build dependencies here?

Possibly. But that is a lot less likely to fail than a compile step -- and perhaps we can ship the signing tool to make it not a dependency. It's different for each platform.

chrisplatorres:
Note from Discord:

~25% ? of times studio loads upside down and without UI
  
 

25Capture

> it's what I think I mentioned before, the CEF mode being unsupported, since we do in-process it's probably a race condition in the boot it especially triggers if you jiggle the mouse, the first thing to try would be to break it out to the supported multiprocess fork

chrisplatorres:
Checks are currently not passing, I see two errors in the [CI log](https://api.travis-ci.org/v3/job/536526376/log.txt):
```
[91m(node:2415) UnhandledPromiseRejectionWarning: ReferenceError: childSyncs is not defined
    at Immediate._topRenderLoop (/app/src/index.js:1136:25)
    at runNextTicks (internal/process/task_queues.js:55:5)
    at processImmediate (internal/timers.js:412:9)
    at process.topLevelDomainCallback (domain.js:126:23)
[0m[91m(node:2415) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:2415) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

...

```
  requestAnimationFrame
    1) raf


  31 passing (15s)
  19 pending
  1 failing

  1) requestAnimationFrame
       raf:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/app/tests/unit/timer/requestAnimationFrame.test.js)
      at Test.Runnable._timeoutError (/app/node_modules/mocha/lib/runnable.js:455:10)
      at Timeout.<anonymous> (/app/node_modules/mocha/lib/runnable.js:266:24)
      at listOnTimeout (internal/timers.js:531:17)
      at processTimers (internal/timers.js:475:7)
```

chrisplatorres:
Checks are passing now.

chrisplatorres:
In `realitytabs.html` when pressing "e" to open the menu:

```
TypeError: Cannot read property '1' of undefined
    at window.addEventListener.e (file:///C:/Users/chris/github/studio/exokit/ex
amples/realitytabs.html:1781:29)
    at emit (events.js:193:13)
    at EventEmitter.emit (domain.js:481:20)
    at _emit (C:\Users\chris\github\studio\exokit\src\Event.js:62:40)
    at window._emit (C:\Users\chris\github\studio\exokit\src\Window.js:1146:28)
    at _emit (C:\Users\chris\github\studio\exokit\src\Event.js:40:14)
    at _recurse (C:\Users\chris\github\studio\exokit\src\Event.js:51:9)
    at _recurse (C:\Users\chris\github\studio\exokit\src\Event.js:55:9)
    at _recurse (C:\Users\chris\github\studio\exokit\src\Event.js:55:9)
    at _recurse (C:\Users\chris\github\studio\exokit\src\Event.js:55:9)
```

modulesio:
☝️ Good catch, sounds gamepads related.

My guess is the gamepads are being pulled from the fake vr display, but they no longer can be with the new global gamepads implementation.

chrisplatorres:
Ok, I am going to log the values and see where there undefined is from.

https://github.com/exokitxr/exokit/blob/studio/examples/realitytabs.html#L1781

chrisplatorres:
In `realitytabs.html` menu when clicking the button to open hterm for an iframe:

```
TypeError: Cannot read property 'nextTick' of undefined
    at module.exports (C:\Users\chris\github\studio\exokit\node_modules\hterm-re
pl\index.js:203:13)
    at Promise (C:\Users\chris\github\studio\exokit\src\DevTools.js:14:7)
    at new Promise (<anonymous>)
    at _getReplServer (C:\Users\chris\github\studio\exokit\src\DevTools.js:12:16
)
    at Object.requestDevTools (C:\Users\chris\github\studio\exokit\src\DevTools.
js:102:30)
    at intersectionCbs.push (file:///C:/Users/chris/github/studio/exokit/example
s/realitytabs.html:2269:53)
    at _updateIntersections (file:///C:/Users/chris/github/studio/exokit/example
s/realitytabs.html:2474:9)
    at animate (file:///C:/Users/chris/github/studio/exokit/examples/realitytabs
.html:2517:3)
    at onAnimationFrame (three.js:21802:36)
    at onAnimationFrame (three.js:7455:4)
```

modulesio:
We'll probably need to not use `process` in [hterm-repl](https://github.com/modulesio/hterm-repl/blob/master/index.js#L203).

That object is deliberately removed from the `window` context and we should not re-add it.

chrisplatorres:
Testing `studio` on quest, 

- `realitytabs.html` UI loads, can click sites
- sites do not appear at all
- can click the iframe move button, but it is empty
- this includes `tutorial.html` that is preloaded
- same with opening a new 3d tab from url bar
- controllers work properly

modulesio:
How does the above compare to `master`?

chrisplatorres:
Master loads sites as expected.

modulesio:
Thanks, this will probably require a deeper debug sesh then...

chrisplatorres:
Similar experience on Magic Leap One.

modulesio:
I think this currently has a bug where hovering the reality tab mover does not correctly hover the right place in the viewport -- perhaps this regressed with the console hiding support?

PR #1004, opened 2019-05-02
Fix tutorial.html texture

chrisplatorres:
Noticed that tutorial.html text becomes unreadable from a distance. 

Before:
blurCapture

After:
slight3Capture


chrisplatorres:
The texture doesn't seem to handle resize correctly.

modulesio:
I don't think it's the texture that needs to handle resize, but the entire page.

PR #1005, opened 2019-05-03
ensure node version is v11.6.0 for android

ngokevin:


chrisplatorres:
Thanks, this would be a good check to have in other places as well such as the ML build script.

modulesio:
For posterity, this is only because `node-gyp` downloads the node header files to use for the build -- which we need to build a compatible native binding.

Other than that, I don't think anything in the scripts cares about node version and going forward I think the right architecture is to include the headers as a module. In which case this restriction would not be necessary.

scripts/build-android.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+NODE_VERSION=`node -v`
+if [ $NODE_VERSION != "v11.6.0" ]
+then
+  echo "[ERROR] Node version needs to be v11.6.0."
+  exit
```
modulesio: Can we exit with an error code like `exit 1`?

Otherwise other tooling might consider this failure to be a successful build.

PR #1021, opened 2019-05-06
Add example and images to Reality Layers iframe doc

chrisplatorres:
This PR documents the `realitytabs.html` example iframe sections on the Reality Layers iframe doc.


modulesio:
Just a heads up, we are probably moving away from this file in favor of the studio.

That file is much too big anyway, and the core functionality should be broken out into more singular examples that are loaded from the studio frontend.

chrisplatorres:
That makes sense, I added this in response to the docs critique:
> Great start, needs to be expanded with images and code examples and Github links.

Is there a better approach for code examples here?

modulesio:
No, I think a simpler iframes layering example needs to be extracted out-- something without menu code getting in the way, just the raw technical details.

This PR is a good start to that, but I think the shape of the doc should be a simplified subset over what's here.

PR #1029, opened 2019-05-07
Add Windows Subsystem for Linux (WSL) requirement in build docs

chrisplatorres:
This PR adds WSL requirement to the Oculus Mobile APK and Magic Leap MPK build instructions in docs.

### DIFF
--- a/docs/guides/exportMagicLeapOneMPK.md
+++ b/docs/guides/exportMagicLeapOneMPK.md
@@ -12,8 +12,9 @@ This guide explains how to package your XR site with the Exokit Engine as an MPK
 In the Exokit [`main.cpp`](https://github.com/exokitxr/exokit/blob/f10dadf0013de0a35a5e72046140a0345987ab80/main.cpp#L416) is where you will find the default site that the Exokit MPK will load into. Change the `jsString` to your site URL or file path.
 
 ## Prerequisites
-Node `11.6.0`
+Node `11.6.0`  
 Latest Exokit
+Ubuntu Windows Subsystem for Linux (WSL)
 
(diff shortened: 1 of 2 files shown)


modulesio:
Did you get this working? I don't think I heard of anyone getting WSL working for android builds, due to Java race conditions.

chrisplatorres:
I will build again and confirm that it worked, if no one else is using WSL for the android build then it is best to keep it out of the build instructions. 🤷‍♂

modulesio:
It would be awesome if it worked -- I think you'd be the first. I remember it randomly failing with copy errors.

If it works, can you document the Java setup you're using? I.e. internal to WSL, external in windows, which JDK, etc.

I think that the only reason we didn't go this route before is because nobody could figure out how to make it build reliably 😓.

chrisplatorres:
Just built again through Ubuntu WSL. 

I think it is using the Java from Windows:  `JAVA_HOME` : jdk-12.0.1

chrisplatorres:
Actually, it is [part of the instructions](https://github.com/exokitxr/exokit/blob/9ff09a4fb4e75ba526276dc7a68e55f88fcd8d71/docs/guides/exportOculusAPK.md#install-android-sdkmanager) to install `default-jdk` in WSL:

```
$ java --version
openjdk 10.0.2 2018-07-17
OpenJDK Runtime Environment (build 10.0.2+13-Ubuntu-1ubuntu0.18.04.4)
OpenJDK 64-Bit Server VM (build 10.0.2+13-Ubuntu-1ubuntu0.18.04.4, mixed mode)
```

PR #1030, opened 2019-05-07
Multithreading test fixes

modulesio:
Experiments in terminating Exokit worker threads cleanly.

Needs major cleanup before merging.

PR #1031, opened 2019-05-07
Multithreaded Worker destruction

modulesio:
In the paralellization branch we weren't destroying worker resources due to dangling native handles.

This PR cleans up those references properly and unlocks worker cleanup while making sure tests pass.

Supercedes https://github.com/exokitxr/exokit/pull/1030.

[WIP]


modulesio:
`<image>` tags parallel loading is still using a [dangling handle](https://github.com/exokitxr/exokit/blob/77a1ca183d498db1831029aad9db01ae0450fdba/deps/exokit-bindings/canvascontext/src/image-context.cc#L86).

Issue #1010, opened 2019-05-04
Linux SteamVR errors


extraymond:
**Describe the bug**
Receive segfault when running exokit installed via the installer script.
 
**To Reproduce**
1.  run exokit cli.

**System information:**
 - OS: ubuntu 18.04
 
 - Exokit Version [e.g. v0.0.512]

 
 - Node version
v11.15.0, installed via nodejs(v11) repo
 
 - Npm version/build log if applicable
 6.7.0

 - Is your Exokit downloaded or built?
Downloaded.

chrisplatorres:
Was able to repro on Ubuntu 18.04 / node v11.5:

```
(gdb) run /usr/local/lib/exokit/src/index.js
Starting program: /home/chris/.nvm/versions/node/v11.15.0/bin/node /usr/local/lib/exokit/src/index.js
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff6a79700 (LWP 22102)]
[New Thread 0x7ffff6278700 (LWP 22103)]
[New Thread 0x7ffff5a77700 (LWP 22104)]
[New Thread 0x7ffff5276700 (LWP 22105)]
[New Thread 0x7ffff4a75700 (LWP 22106)]
[New Thread 0x7ffff7ff6700 (LWP 22107)]
internal/modules/cjs/loader.js:857
  return process.dlopen(module, path.toNamespacedPath(filename));
                 ^

Error: libopenvr_api.so: cannot open shared object file: No such file or directory
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:857:18)
    at Module.load (internal/modules/cjs/loader.js:685:32)
    at Function.Module._load (internal/modules/cjs/loader.js:620:12)
    at Module.require (internal/modules/cjs/loader.js:723:19)
    at require (internal/modules/cjs/helpers.js:14:16)
    at /usr/local/lib/exokit/src/native-bindings.js:19:18
    at Object.<anonymous> (/usr/local/lib/exokit/src/native-bindings.js:26:3)
    at Module._compile (internal/modules/cjs/loader.js:816:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:827:10)
    at Module.load (internal/modules/cjs/loader.js:685:32)
[Thread 0x7ffff4a75700 (LWP 22106) exited]
[Thread 0x7ffff5276700 (LWP 22105) exited]
[...]

PR #1033, opened 2019-05-08
Headset metrics blitting

modulesio:
Fixes https://github.com/exokitxr/exokit/issues/1010.

The blitting math in the `window` compose layers implementation was not accurate to the headset framebuffer size.


modulesio:
Tested on MacOS with https://github.com/exokitxr/exokit/pull/1033/commits/fb4606ee47d431d01721e50b93f24592a0a03538 and it seems to work:

image

The related fix in `master` is https://github.com/exokitxr/exokit/pull/1036.

PR #1040, opened 2019-05-09
Multithreaded WebAudio

modulesio:
Fixes per-Worker WebAudio for multithreaded reality tabs.


modulesio:
Tested this on Oculus Go and it passes tests, so let's continue in `user-parallel`.

ngokevin:
thanks!

PR #1048, opened 2019-05-10
Window URL navigation

modulesio:
Exokit single-threaded supported location for a long time, but the multithreaded windowing model threw a wrench in URL navigation support, since that requires a dance between the parent and child context.

This PR adds `window.location` navigation support (via all of the usual `window.location` setters) to the mulithreaded `window` case.

This PR should support the usual XR navigation cases where a window chooses to navigate, as well as transparent navigation retargeting for reality tabs without losing the tab.


modulesio:
Test site: https://a-frobot.github.io/aframe/examples/showcase/link-traversal/

PR #1051, opened 2019-05-11
Optimize wait get poses

modulesio:
This abstracts the enormous update functions in `index.js`, which is our main render loop.

This is to make them more conducive to hot path optimization and reduce GC pressure in the main critical code path of our code.


modulesio:
This needs a quick rebase on top of new the WebXR extensions in `studio`.

modulesio:
Rebased

PR #1052, opened 2019-05-11
WebXR hardware extensions APIs

modulesio:
This PR refactors the non-standard WebXR hardware bindings (meshing, planes, hands, eyes) into a generic system that is not tied to the Magic Leap API binding.

The reasons for doing this are:

- Simpler to manage the bindings in one place at the top level render loop, instead of internally in the user windows
- We now emulate these APIs in the studio, allowing us to test and develop against them without needing hardware
- Allows us to support more hardware (such as Leap Motion/Zed/Valve Index) without leaking hardware API warts specific to a particular platform
- More standards-compliant -- these are finding their way into the WebXR spec and we want Exokit to follow suit

### DIFF
--- a/deps/exokit-bindings/magicleap/src/magicleap.cc
+++ b/deps/exokit-bindings/magicleap/src/magicleap.cc
@@ -451,7 +451,7 @@ NAN_METHOD(MLRaycaster::WaitGetPoses) {
       MLMat4f hitMatrix = composeMatrix(position, quaternion, scale);
 
       Local<Object> xrHitResult = Nan::New<Object>();
-      Local<Float32Array> hitMatrixArray = Float32Array::New(ArrayBuffer::New(Isolate::GetCurrent(), (void *)hitMatrix.matrix_colmajor, 16 * sizeof(float)), 0, 16);
+      Local<Float32Array> hitMatrixArray = Float32Array::New(ArrayBuffer::New(Isolate::GetCurrent(), (void *)hitMatrix.matrix_colmajor, sizeof(hitMatrix.matrix_colmajor)), 0, 16);
       xrHitResult->Set(JS_STR("hitMatrix"), hitMatrixArray);
       Local<Array> result = Nan::New<Array>(1);
       result->Set(0, xrHitResult);
@@ -564,8 +564,8 @@ MLMesher::MLMesher(float range, MLMeshingLOD lod) : range(range), lod(lod), mesh
 MLMesher::~MLMesher() {}
 
 NAN_METHOD(MLMesher::New) {
-  int lodValue = TO_INT32(info[0]);
-  float range = TO_FLOAT(info[1]);
+  float range = TO_FLOAT(info[0]);
+  int lodValue = TO_INT32(info[1]);
   
   MLMeshingLOD lod;
   switch (lodValue) {
@@ -650,8 +650,8 @@ NAN_METHOD(MLMesher::WaitGetPoses) {
       // std::unique_lock<std::mutex> lock(mlContext->positionMutex);
 
       meshExtents.center = mlContext->position;
-      // meshExtents.rotation =  mlContext->rotation;
-      meshExtents.rotation = {0, 0, 0, 1};
+      meshExtents.rotation =  mlContext->rotation;
+      // meshExtents.rotation = {0, 0, 0, 1};
     }
     meshExtents.extents.x = mlMesher->range;
     meshExtents.extents.y = mlMesher->range;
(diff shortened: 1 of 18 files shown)


modulesio:
Note that the examples have all been updated to use the new, more standard APIs for these extensions:

```
      const session = await display.requestSession({
        exclusive: true,
        extensions: {
          meshing: true,
        },
      });
        session.addEventListener('meshadd', _meshadd);
        session.addEventListener('meshupdate', _meshupdate);
        session.addEventListener('meshremove', _meshremove);
```

Also, these examples are all available to test with emulated data in the studio.

For example, we will generate some fake cubes nearby for the meshing example, we will generate some planes for the planes example. And the hands example will generate a freakish hand model to test that the bones data can be rendered.

modulesio:
We still need to port the actual magic leap APIs to be based on the new system, but that's mostly just moving code around since the bindings already exist.

src/VR.js
```diff
@@ -356,23 +354,10 @@ class VRDisplay extends EventEmitter {
...
-      for (const k in init.detail) {
-        this[k] = init.detail[k];
-      }
-    }
-  }
-}
-
-class FakeMesher extends EventTarget {
-  constructor(session) {
+class FakeMesher extends EventEmitter {
```
modulesio: We should move all of the VR fakers to a separate `FakeXR.js` so it doesn't get in the way of the standard binding code.

We could also probably not even `require` it in the main engine case, where we don't use fake XR devices at all.

src/index.js
```diff
@@ -197,7 +197,7 @@ const xrState = (() => {
       return result;
     };
   };
-  const _makeTypedArray = _makeSab(4*1024);
+  const _makeTypedArray = _makeSab(8*1024);
```
chrisplatorres: 8*1024 as opposed to 8192 for legibility?

src/index.js
```diff
@@ -294,154 +326,145 @@ const topVrPresentState = {
...
-    events: 0,
-    media: 0,
-    user: 0,
-    submit: 0,
-    total: 0,
-  };
-  const TIMESTAMP_FRAMES = 100;
+const requestPresent = async () => {
+  const hmdType = getHMDType();
+  console.log('request present', hmdType); // XXX
```
chrisplatorres: I've noticed the `// XXX` comment in different areas, what does it signify?

src/index.js
```diff
@@ -918,6 +941,112 @@ const _startTopRenderLoop = () => {
...
+              }
+            }
+          }
+        }
+      };
+      _updateHandTracking();
+      
+      const _updateEyeTracking = () => {
+        if (xrState.eyeTracking[0]) {
+          const blink = (Date.now() % 2000) < 200;
```
chrisplatorres: This is used to randomly generate blinks?

src/VR.js
chrisplatorres: I agree, it seems out of place to include the fakes in the core engine (`index.js`, etc.).

chrisplatorres:
Started updating the ["WebXR Hardware Extension APIs docs"](https://exokit.org/docs/api/webxrHardwareExtensionAPIs.html) at #1053 based on this PR's refactor.

src/index.js
modulesio: We needed this extra memory space because we added more fields to the shared xt state.

src/index.js
modulesio: I use it as a todo note, often a log to consider removing.

src/index.js
modulesio: Correct!

src/index.js
modulesio: But yes, it probably wouldn’t matter that this is exactly 8k; it’s just easier to reason about.

Typed arrays also have alignment requirements about, so using non-round values could lead up useless memory in the fragmentation. But that’s just nitpicking for this case.

modulesio:
This PR now has the magic leap hardware bindings to the new WebXR extension APIs.

modulesio:
The main world-aware APIs should be ported and working on Magic Leap now.

I still need to clean up some thousand-line commits.

deps/exokit-bindings/magicleap/src/magicleap.cc
```diff
@@ -844,134 +940,248 @@ NAN_SETTER(MLPlaneTracker::OnPlanesSetter) {
...
+  if (MLHandTrackingCreate(&tracker) != MLResult_Ok) {
+    ML_LOG(Error, "%s: Failed to create hand tracker.", application_name);
+  }
+
+  MLHandTrackingConfiguration handTrackingConfig;
+  handTrackingConfig.handtracking_pipeline_enabled = true;
+  handTrackingConfig.keypoints_filter_level = MLKeypointFilterLevel_1;
+  handTrackingConfig.pose_filter_level = MLPoseFilterLevel_1;
+  for (int i = 0; i < MLHandTrackingKeyPose_Count; i++) {
+    handTrackingConfig.keypose_config[i] = true;
```
modulesio: Note for posterity: this flag set actually changes the way hand bones are detected. If you filter out poses, the bones will get locked in to a static coordinate frame and other weirdness.

Sadly, this is not documented anywhere in the Magic Leap SDK,

modulesio:
Folding this into `studio`.

PR #1060, opened 2019-05-15
Update node versions 11.6.0 -> 12.0.0

chrisplatorres:
This PR updates any docs and paths that are node version `11.6.0` to `12.0.0`. 

This is for the `user-parallel` #760 and #877 move to [Node 12](https://nodejs.org/ko/blog/release/v12.0.0/).


modulesio:
This build is failing -- we should base it off of `here`: https://github.com/exokitxr/exokit/pull/1064

modulesio:
Also, we are moving to `node` `12.2.0`, not strictly `12`.

modulesio:
The way it shook out is the android/magic leap version will be `11.15.0` for now (required during build).

For desktop, however, it can be any version of `12`. Either way the `--experimental-worker` flag should no longer be required.

chrisplatorres:
Ok I will update the android/magicleap cases.

chrisplatorres:
Going to rebase this off `user-parallel`

Issue #1025, opened 2019-05-07
A-Painter 404s when URL does not have a trailing slash


chrisplatorres:
**Describe the bug**
What happened, and what did you expect to happen?
 
running a-painter with a trailing slash as a tab - `node --experimental-worker . -h -t https://aframe.io/a-painter/` gives a normal log.

running a-painter without a trailing slash in URL as a tab, giving a lot of 404s - `node --experimental-worker . -h -t https://aframe.io/a-painter`
 

**Additional context**

I think a-painter treats it different when there is a trailing slash or not.

**System information:**

Windows 10
Built from source
`user-parallel`
Node v11.6.0
Npm v 6.5.0-next.0

modulesio:
Without the `/`:

```
ErrorEvent {
  type: 'error',
  target: <script src="vendor/aframe-master.min.js"></script>,
  bubbles: false,
  cancelable: false,
  defaultPrevented: false,
  propagationStopped: false,
  currentTarget: <script src="vendor/aframe-master.min.js"></script>,
  message:
   'script src got invalid status code: 404 : vendor/aframe-master.min.js',
  stack:
   'Error: script src got invalid status code: 404 : vendor/aframe-master.min.js\n    at ownerDocument.defaultView.fetch.then.res (C:\\Users\\avaer\\Documents\\GitHub\\exokit\\src\\DOM.js:1700:35)\n    at process.internalTickCallback (internal/process/next_tick.js:77:7)' }
Error: script src got invalid status code: 404 : vendor/aframe-master.min.js
    at ownerDocument.defaultView.fetch.then.res (C:\Users\avaer\Documents\GitHub\exokit\src\DOM.js:1700:35)
    at process.internalTickCallback (internal/process/next_tick.js:77:7)
ErrorEvent {
  type: 'error',
  target:
   <script src="vendor/aframe-input-mapping-component.js"></script>,
  bubbles: false,
  cancelable: false,
  defaultPrevented: false,
  propagationStopped: false,
  currentTarget:
   <script src="vendor/aframe-input-mapping-component.js"></script>,
  message:
   'script src got invalid status code: 404 : vendor/aframe-input-mapping-component.js',
  stack:
[...]

PR #1061, opened 2019-05-16
Uncomment trailing slash check in user-parallel

chrisplatorres:
This PR is to uncomment trailing slash check in user-parallel.


chrisplatorres:
Closes #1025

PR #1062, opened 2019-05-16
Disable right click menu in Studio

chrisplatorres:
The PR returns false oncontextmenu for launcher's index.html which disables the annoying right click menu.
  
rightclickCapture

### DIFF
--- a/examples/launcher/src/App.js
+++ b/examples/launcher/src/App.js
@@ -2,6 +2,9 @@ import React from 'react';
 import './css/App.css';
 import Engine from './components/Engine.jsx'
 
+document.oncontextmenu = function() {
+    return false;
+}
 function App() {
   return (
     <div className="App" id="app">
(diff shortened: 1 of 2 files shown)


modulesio:
This looks good, but is it possible to put this into `App.js` and bind it with React?

Going forward we'll probably want a custom context menu there rather than a global hijack.

examples/launcher/src/App.js
```diff
@@ -2,6 +2,9 @@ import React from 'react';
 import './css/App.css';
 import Engine from './components/Engine.jsx'
 
+document.oncontextmenu = function() {
```
modulesio: I was thinking `onContextMenu={() => false}` in the JSX.

We just need to test whether that works as advertised.

chrisplatorres: Testing now.

chrisplatorres: moved in cfe0616

PR #1063, opened 2019-05-16
[WIP] Studio exit

chrisplatorres:
This PR handles exiting Exokit when user clicks on "Exit" from dropdown menu in Studio.

xitCapture


examples/launcher/src/components/Engine.jsx
```diff
@@ -106,14 +106,20 @@ class Engine extends React.Component {
...
+
     openMenu(item) {
       const open = this.state.item !== item;
       this.setState({item: open ? item : null}, () => {
         this.postMenuStatus();
       });
     }
-    
+
+    exitExokit() {
```
chrisplatorres: Should this be a different name? `exitStudio()` ?

examples/studio.html
```diff
@@ -905,6 +905,14 @@
           viewport.set(m.data.viewport[0], 1 - m.data.viewport[1] - m.data.viewport[3], m.data.viewport[2], m.data.viewport[3]);
           break;
         }
+        case 'exit': {
+          context.destroy();
+          window.document.resources.addEventListener('drain', () => {
+            console.log('drain');
+            process.exit();
```
modulesio: `process` should really not be available here. If it is, that is a separate problem.

examples/launcher/src/components/Engine.jsx
modulesio: Probably `exitStudio`.

examples/studio.html
chrisplatorres: Oh no, process definitely isn't available. I just noticed I crashed my way to an exit here.

modulesio: Phew!

chrisplatorres:
Note from Discord:

~25% ? of times studio loads upside down and without UI
25Capture

> it's what I think I mentioned before, the CEF mode being unsupported, since we do in-process it's probably a race condition in the boot it especially triggers if you jiggle the mouse, the first thing to try would be to break it out to the supported multiprocess fork

chrisplatorres:
Going to close this PR to port over to seperate studio repo.

modulesio:
Can we keep it open until that happens, so we don't forget? It's easier for closed PR's to get lost...

chrisplatorres:
Closing now that it is copied in https://github.com/exokitxr/studio/issues/2.

PR #1065, opened 2019-05-17
Studio transform bug

chrisplatorres:
This PR fixes a studio bug where when you click and hold transform while moving mouse outside of window, let go of click, moving cursor in viewport still is in transform mode.

### DIFF
--- a/examples/studio.html
+++ b/examples/studio.html
@@ -910,6 +910,9 @@ const {
           if (target === 'fakeXr') {
             renderer.domElement.requestPointerLock();
             _bindEventTarget('xr');
+          } else if (target === 'engineRender') {
+            console.log("engineRender targeted");
+            _bindEventTarget('engineRender');
           } else if (target === 'xr') {
             console.log('xr click');
           } else if (target === 'transform') {
@@ -1095,11 +1098,14 @@ const _bindEventTarget = target => {
 
           transformControls.addEventListener('dragging-changed', e => {
             orbitControls.enabled = !e.value;
+            transformEnabled = true;
           });
           transformControls.addEventListener('change', e => {
-            iframe.position = mesh.position.toArray();
-            iframe.orientation = mesh.quaternion.toArray();
-            iframe.scale = mesh.scale.toArray();
+            if (transformEnabled) {
+              iframe.position = mesh.position.toArray();
+              iframe.orientation = mesh.quaternion.toArray();
+              iframe.scale = mesh.scale.toArray();
+            }
           });
 
           const mesh = new THREE.Object3D();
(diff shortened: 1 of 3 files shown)


chrisplatorres:
there is a hack here with `transformEnable` that disables the transform movement when the mouse enters the UI, but that does not cover 100% cases: this bug still occurs when dragging mouse on the left side where there is no UI:

leftCapture

chrisplatorres:
The next step is detecting when the users mouse is no longer moving in engine-render https://github.com/exokitxr/exokit/pull/1065/files#diff-818c6db2ba21b9def5bafb0e208c1111R215

modulesio:
Is there a way to get `mouseup` when not in the window? I figure GLFW must have that capability 🤔

chrisplatorres:
Not finding anything in a quick search, should dragging outside of the window be allowed behavior here?

modulesio:
Yes, I think so. We just need to handle the mouse moving up.

chrisplatorres:
`mouseup` is now handled on `window`, which fixes this bug.

examples/studio.html
```diff
@@ -159,6 +159,9 @@
 const localLine = new THREE.Line3();
 const localRaycaster = new THREE.Raycaster();
 
+window.addEventListener('mouseup', e => {
```
chrisplatorres: This might need to be moved to a more appropriate location

modulesio: I think the handler is fine, but the way the transform controls work is that we create/destroy the object when the controls appear.

I think we should unify under a single path rather than add a new mode for `transformEnabled`. I can't convince myself this is free of corner cases otherwise.

Does calling `_removeTransformControls` work instead?

chrisplatorres: What happens in that case is the `transformControls` are removed, but the iframe still moves when the mouse cursor is moved back in the viewport.

modulesio: Hm, if it's the transform controls that are doing that, and destroying them does not disable them, then I think that's a bug in the transform controls destruction/timing.

I would look into why the events continue to be handled for a dead transform controls in that case.

modulesio: Code pointer: https://github.com/exokitxr/exokit/pull/1065/files#diff-5f6b5375e965927cdac457568760b24aR1124

chrisplatorres: got it:

- studio.html disposes of transformControls - https://github.com/exokitxr/exokit/blob/studio/examples/studio.html#L1118
-  disposing of transformControls removes mouseup listener - https://github.com/exokitxr/exokit/blob/studio/examples/TransformControls.js#L126
- mouseup listener is the only way that would document.removeEventListener( "mousemove", onPointerMove, false );

so simply adding to the dispose in TransformControls.js:
document.removeEventListener( "mousemove", onPointerMove, false );

fixes it

PR #1071, opened 2019-05-19
[Go/Quest] Direct blit

modulesio:
This PR drops the blit from window texture to the OVR swapchain from the Oculus Go/Quest render pipeline.

The trick to doing this was to multiply the tan angle texture matrix to allow a single side-by-side texture to be used for both eyes.

This makes the Oculus Go render pipeline basically the same as the Oculus desktop one, with incrementing swap chain passed to the user window and no finalization blit.

Also cleans up some Oculus desktop binding.

Fixes https://github.com/exokitxr/exokit/pull/1071.
Supercedes https://github.com/exokitxr/exokit/pull/930.


modulesio:
Trying a GLSync solution here to address the frame skipping issue.

dmarcos:
Nice. Thanks for doing this

modulesio:
Continuing work in `user-parallel`.

Issue #1035, opened 2019-05-08
Update windows code signing certificate


modulesio:
Our windows (DigiCert) signing certificate is due to expire next week, so we'll need to renew for the installer flow to continue working.

chrisplatorres:
Also the current publisher listed in cert is: "Zeo Incorporated" 

![zeoincwindows](https://user-images.githubusercontent.com/29695350/57412829-c20b8700-71b7-11e9-8fef-8e4293c48dd3.PNG)

modulesio:
Yeah, we'd change that as part of this.

modulesio:
Put in for this one.

PR #1072, opened 2019-05-19
Windows verified installer

modulesio:
This PR adds a verified publisher certificate (under `Exokit Inc`) for the windows installer. The certificate is encrypted in this repo.

Fixes https://github.com/exokitxr/exokit/issues/1035.


modulesio:
The newly signed installer should ship with the next version of Exokit 🦄.

PR #1073, opened 2019-05-19
Simpler Studio layer view

chrisplatorres:
This PR replaces the DOM view with a simpler layer view.

### DIFF
--- a/examples/launcher/src/components/Dom.jsx
+++ b/examples/launcher/src/components/Dom.jsx
@@ -40,8 +40,6 @@ class Dom extends React.Component {
           attrs: [{
             name: 'lol',
             value: 'zol',
-            name: 'woot',
-            value: 'toot',
           }],
           childNodes: [],
         },
@@ -99,7 +97,6 @@ class Dom extends React.Component {
     return (
       <div className="Dom">
         <DomList root={this.state.root} selectEl={this.state.selectEl} onClick={el => this.onClick(el)} />
-        {this.state.selectEl ? <DomDetail el={this.state.selectEl} epoch={this.state.epoch} /> : null}
       </div>
     );
   }
@@ -145,6 +142,7 @@ class DomItem extends React.Component {
 
     this.state = {
       open: true,
+      dropdownOpen: false,
     };
   }
 
@@ -189,6 +187,42 @@ class DomItem extends React.Component {
     e.stopPropagation();
   }
 
+  toggleDropdownOpen(e) {
+    this.setState({
+      dropdownOpen: !this.state.dropdownOpen,
+    });
+
+  }
+
+  cloneTab() {
+    const {el} = this.props;
+    let url;
+
+    el.attrs.find(attr => (attr.name === "src") ? url = attr.value : null);
+
+    window.postMessage({
+      method: 'open',
+      url,
+      d: 3,
+    });
+
+    this.toggleDropdownOpen(el);
+  }
+
+  deleteTab() {
+    const {el} = this.props;
+
+    window.postMessage({
+      method: 'edit',
+      keypath: el.keypath,
+      edit: {
+        type: 'remove',
+      },
+    });
+
+    this.toggleDropdownOpen(el);
+  }
+
   render() {
     const {el, level} = this.props;
 
@@ -196,10 +230,16 @@ class DomItem extends React.Component {
       return (
         <li className={this.getClassnames()}>
           <div className="dom-item-label" style={this.getStyle()} onClick={() => this.props.onClick(el)} onMouseEnter={() => this.props.onMouseEnter(el)} onMouseLeave={() => this.props.onMouseLeave(el)} ref={this.domElRef} tabIndex={-1}>
-            <div className="dom-item-arrow" onClick={e => this.toggleOpen(e)}>...</div>
+            <div className="dom-item-arrow" onClick={e => this.toggleDropdownOpen(e)}>...</div>
             <div className="dom-item-name"> {_el2Text(el)}</div>
           </div>
-          </li>
+          <div className="dom-item-dropmenu">{this.state.dropdownOpen ?
+            <div className="dom-detail">
+              <div className="dom-detail-button" onClick={() => this.cloneTab()}>Clone</div>
+              <div className="dom-detail-button" onClick={() => this.deleteTab()}>Delete</div>
+            </div>
+             : null}</div>
+        </li>
         );
     } else if (el.nodeType === Node.ELEMENT_NODE) {
       return (
@@ -225,10 +265,11 @@ class DomItem extends React.Component {
   }
 }
 
+/*
 class DomDetail extends React.Component {
-  /* constructor(props) {
+   constructor(props) {
     super(props);
-  } */
+  }
 
   render() {
     const {el} = this.props;
@@ -316,5 +357,6 @@ class DomAttribute extends React.Component {
     );
   }
 }
+*/
 
 export default Dom;
--- a/examples/launcher/src/components/Dom.jsx
+++ b/examples/launcher/src/components/Dom.jsx
@@ -8,6 +8,15 @@ const _decorateKeypaths = (el, keypath = []) => {
   }
 };
 const _el2Text = el => {
+  let result = '<' + el.tagName.toLowerCase();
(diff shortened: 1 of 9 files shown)


examples/launcher/src/components/Engine.jsx
```diff
@@ -187,7 +187,7 @@ class Engine extends React.Component {
       const url = urlInput.value;
 
       window.postMessage({
-        method: 'open',
+        method: 'reload',
         url,
         d: null,
```
modulesio: What dimensionality should the new tab be? I'm guessing `3`.

chrisplatorres: Perhaps there should be 2 separate "reload buttons":

- Close all tabs, and open a 2d layer 
- Close all layers, and open a 3d layer

one for d of `2` and one for `3`

modulesio: If the intent of this functionality is to be "like a browser" then we probably don't want any additional knobs here...

chrisplatorres: 2d3dCapture

It is mirroring the current options of opening a new 2d/3d layer^

modulesio: But what will the effect be on `openUrl` when no dimension is specified?

chrisplatorres: `openUrl` defaults to d of `3` https://github.com/exokitxr/exokit/blob/studio/examples/studio.html#L608

modulesio: I see, so this is just confusing code then. Either way I'm not sure if the reload concept makes any sense, because it's not clear what the UX of it would be compared to the other cases.

It seems we're navigating a different abstraction here -- we are navigating an orchestration of realities, not a reality tab.

Maybe what we are looking for is the ability to edit the src of a reality tab?

chrisplatorres: I think the right thing for this case is "Clear all layers", which would use this current PR with no opening of a new url.

modulesio: In general, a stupider way to navigate iframes might be best, then? Think the DOM view, except only reality tabs, with a clear button.

chrisplatorres: Generally the most front-facing UI should be very stupid, the DOM view is even too hairy right now,

> DOM view, except only reality tabs, with a clear button. 
 
I think this should be it, with the ability to click into and *then* see/edit the full DOM/source

modulesio: +1 to that.

In that case we wouldn't need the reload option at all.

chrisplatorres:
Current "layer view", you click `...` then there is a dropdown menu that has:
- clone/duplicate layer
- delete layer
- see DOM view 
- ?edit layer in text editor?

  
layr4Untitled

examples/launcher/src/components/Engine.jsx
chrisplatorres: removed in f9febc4

chrisplatorres:
stuguntCapture

reference style picture

examples/launcher/src/components/Dom.jsx
```diff
@@ -8,12 +8,15 @@ const _decorateKeypaths = (el, keypath = []) => {
   }
 };
 const _el2Text = el => {
-  let result = '<' + el.tagName.toLowerCase();
-  for (let i = 0; i < el.attrs.length; i++) {
-    const attr = el.attrs[i];
-    result += ' ' + attr.name + '=' + JSON.stringify(attr.value);
+  let result = '';
```
modulesio: Isn't this deleting the old functionality, gutting the DOM view entirely?

examples/launcher/src/components/Dom.jsx
```diff
@@ -37,8 +40,6 @@ class Dom extends React.Component {
           attrs: [{
             name: 'lol',
             value: 'zol',
-            name: 'woot',
```
modulesio: Good catch!

examples/launcher/src/components/Engine.jsx
```diff
@@ -204,6 +192,16 @@ class Engine extends React.Component {
       this.blur();
     }
 
+    postViewportMessage() {
```
modulesio: This code is still duplicated [here](https://github.com/exokitxr/exokit/pull/1073/files#diff-818c6db2ba21b9def5bafb0e208c1111R26).

chrisplatorres: c86b1c4

examples/launcher/src/components/Dom.jsx
chrisplatorres: Yes, I'll revert that since DOM view is going to be used later on.

chrisplatorres: aeafc1f

chrisplatorres:
CI failing on tests for mac/linux

```
  31 passing (15s)
  19 pending
  1 failing

  1) requestAnimationFrame
       raf:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/app/tests/unit/timer/requestAnimationFrame.test.js)
      at Test.Runnable._timeoutError (/app/node_modules/mocha/lib/runnable.js:455:10)
      at Timeout.<anonymous> (/app/node_modules/mocha/lib/runnable.js:266:24)
      at listOnTimeout (internal/timers.js:531:17)
      at processTimers (internal/timers.js:475:7)
```

modulesio:
☝️ The tick loop is probably not ticking in that case.

chrisplatorres:
It looks like it is inheriting the same error as the `studio` branch

modulesio:
Thanks, there were a bunch of conflicting merges there when we shipped parallelization (still ongoing conflicts).

Is this a blocker for you?

chrisplatorres:
It's fine, I will make separate PRs and merge them into this branch for now.

examples/launcher/src/components/Dom.jsx
modulesio: Looks good 👍

examples/launcher/src/components/Engine.jsx
```diff
@@ -1,8 +1,19 @@
 import React from 'react';
+import Resizable from 're-resizable';
 import Dom from './Dom';
 import Console from './Console';
 import '../css/engine.css';
 
+const postViewportMessage = () => {
```
modulesio: This is a nitpick, but in case it's not clear, we generally prefix "private" methods with `_` in the codebase.

"Private" means "not exported out of this scope". In lieu of a true language feature for it, it makes it easier to tell at a glance the intent of the code.

examples/launcher/src/components/Engine.jsx
```diff
@@ -220,6 +208,17 @@ class Engine extends React.Component {
       });
     }
 
+    onConsoleClick() {
+      let studioConsole = document.getElementById('console');
+      if(studioConsole.clientHeight === 0){
+        studioConsole.style.height = "100px";
```
modulesio: This is better done with a `.open` class + css.

Once we start getting into the game of sizing things in the Javascript it's prone to quickly devolve into nightmare math.

examples/launcher/src/components/Engine.jsx
chrisplatorres: e447ac0

examples/launcher/src/components/Engine.jsx
chrisplatorres: b7eeba5 61b6e79 99af7bf

examples/launcher/src/components/Dom.jsx
```diff
@@ -171,41 +181,88 @@ class DomItem extends React.Component {
...
+      dropdownOpen: !this.state.dropdownOpen,
+    });
+
+  }
+
+  cloneTab() {
+    const {el} = this.props;
+    let url;
+
+    el.attrs.find(attr => (attr.name === "src") ? url = attr.value : null);
```
modulesio: Maybe:

```
const url = el.attrs.find(attr => attr.name === 'src').value;
```

chrisplatorres: 12b3f46

PR #1076, opened 2019-05-21
[WIP] Studio resizable UI

chrisplatorres:
This PR add resizing functionality to the Studio UI.

draggy

### DIFF
--- a/examples/launcher/src/components/Dom.jsx
+++ b/examples/launcher/src/components/Dom.jsx
@@ -265,6 +265,59 @@ class DomItem extends React.Component {
   }
 }
 
+class DomDetail extends React.Component {
+  /* constructor(props) {
+    super(props);
+  } */
+  toggleOpen() {
+    const {el} = this.props;
+
+    this.setState({
+      open: !this.state.open,
+    });
+
+    el.stopPropagation();
+  }
+
+  cloneTab() {
+    const {el} = this.props;
+    let url;
+
+    el.attrs.map(attr => (attr.name === "src") ? url = attr.value : null);
+
+    window.postMessage({
+      method: 'open',
+      url,
+      d: 3,
+    });
+
+    this.toggleOpen();
+  }
+
+  deleteTab() {
+    const {el} = this.props;
+
+    window.postMessage({
+      method: 'edit',
+      keypath: el.keypath,
+      edit: {
+        type: 'remove',
+      },
+    });
+
+    this.toggleOpen();
+  }
+  render() {
+    const {el} = this.props;
+
+    return (
+      <div className="dom-detail">
+        <div className="dom-detail-button" onClick={() => this.cloneTab()}>Clone</div>
+        <div className="dom-detail-button" onClick={() => this.deleteTab()}>Delete</div>
+      </div>
+    );
+  }
+}
 
 class DomAttribute extends React.Component {
   constructor(props) {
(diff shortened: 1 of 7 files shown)


chrisplatorres:
There are a few things that need to be done here still:
- resizing one menu also updates other UI sizes
- should more of the UI be resizable? It seems only the right sidebar needs to be resized at the moment

modulesio:
> should more of the UI be resizable? It seems only the right sidebar needs to be resized at the moment

Probably the bottom console bar too, though it should also have a hide ability.

chrisplatorres:
> Probably the bottom console bar too, though it should also have a hide ability.

I'll add that into this PR, the console will be hidden by default, but when it is visible it is resizable.

chrisplatorres:
Cloning and deleting tabs, as well as resizing works as expected. Going to fix up the UI so that..

click `...`
then there is a  menu similar to above attr one, but is smaller, directly under the site, and has options for:
- clone 
- delete
- edit (auto-detect site and open in an three/aframe/babylon editor or if not able to be detected, open in a text editor)

modulesio:
Awesome!

The editor break-out/auto-detect sounds like it's much bigger than the rest. Did you want to merge this one first and then continue with a new PR?

chrisplatorres:
Yes, and I agree.. the "editing a site" part of Studio should be in a separate PR.

chrisplatorres:
This can now be merged and will continue work through other branches.

st06Capture

examples/launcher/src/components/Dom.jsx
```diff
@@ -229,14 +269,51 @@ class DomDetail extends React.Component {
   /* constructor(props) {
     super(props);
   } */
+  toggleOpen() {
```
modulesio: If [`DomDetail` is removed](https://github.com/exokitxr/exokit/pull/1076/files#diff-43a85fe06ede43ea8d3091dd359ef2b7L102), why is this being edited as well?

examples/launcher/src/components/Dom.jsx
```diff
@@ -99,7 +97,6 @@ class Dom extends React.Component {
     return (
       <div className="Dom">
         <DomList root={this.state.root} selectEl={this.state.selectEl} onClick={el => this.onClick(el)} />
```
modulesio: If the intent is to re-add `DomDetail` later, perhaps just comment it?

examples/launcher/src/components/Dom.jsx
```diff
@@ -189,17 +187,59 @@ class DomItem extends React.Component {
...
+      dropdownOpen: !this.state.dropdownOpen,
+    });
+
+  }
+
+  cloneTab() {
+    const {el} = this.props;
+    let url;
+
+    el.attrs.map(attr => (attr.name === "src") ? url = attr.value : null);
```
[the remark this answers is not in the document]
chrisplatorres: b5d4ae4

examples/launcher/src/components/Dom.jsx
chrisplatorres: 29588b7

examples/launcher/src/components/Dom.jsx
chrisplatorres: reverted in 
ad9ecad

chrisplatorres:
Screenshot for style referencing:
stu1Capture

PR #1077, opened 2019-05-21
Fix

shawwn:


shawwn:
Actually, I guess this isn't necessary. It would only matter if the function created a handle scope, which it doesn't.

PR #1079, opened 2019-05-21
Remove appveyor artifacts config

modulesio:
Appveyor recently instituted new limits on artifacts size, which we hit.

We only use Github artifacts anyway for releases, so this PR removes the artifact config from `appveyor.yml`.


modulesio:
[Passing on AppVeyor](https://ci.appveyor.com/project/modulesio/exokit/builds/24688745)

PR #1084, opened 2019-05-21
Escape pipe in table columns

OnlineCop:
Replaces escaped pipe with HTML entity 124: &#124;.
Replaces double hyphens inside the table with HTML entity 8209 (non-breaking hyphen): &#8209;.

This looks ugly/wrong in the "Preview changes" diff so it may need to be tested locally to ensure this doesn't break anything visually.

(Previous fix for this was issue #1083 which looked correct in "Preview changes" but still rendered incorrectly when published.)


chrisplatorres:
> This looks ugly/wrong in the "Preview changes" diff so it may need to be tested locally to ensure this doesn't break anything visually.

Thanks, this will need to be tested locally to make sure it looks as expected.

chrisplatorres:
Output of this code locally (ignore font bold weirdness):

012Capture

OnlineCop:
Yeah; that's definitely not going to work. Someone in Discord suggested the &#8209; (non-breaking hyphens) to replace the double hyphens, so that can easily go.

I was hoping that it would at least fix the pipe within the `--webgl <1|2>` text, but as you can see, it's outputting `&#124;`.

https://stackoverflow.com/questions/17319940/how-to-escape-a-pipe-char-in-a-code-statement-in-a-markdown-table

Do any of you have ideas as to how the pipe can be used there? If not, then the format may need to change to be something like `--webgl <1,2>` (comma instead of a pipe), if that will be the only way to get the table to render properly.

OnlineCop:
As you saw with #1083, escaping the pipe (my first idea) was one of the suggestions on that SO link, but that doesn't appear to have fixed the issue.

Is the markdown language version possibly old? Is it a different form of markdown where a wiki somewhere could explain the correct way to do it?

modulesio:
GitHub renders it equally weird in the preview.

Capture 26

modulesio:
@OnlineCop accidentally closed this one with a force-push; feel free to re-open!

PR #1085, opened 2019-05-21
Minor punctuation fixes

OnlineCop:
Adds colons to text immediately above user-typed lines.
Replaces 2nd 'is' with 'it' in the text "is that is will".

### DIFF
--- a/docs/guides/exportOculusAPK.md
+++ b/docs/guides/exportOculusAPK.md
@@ -92,7 +92,7 @@ To install NDK directly, run:
 
 ### Build APK
 
-Confirm that `$ANDROID_HOME` points to Android/SDK directory
+Confirm that `$ANDROID_HOME` points to Android/Sdk directory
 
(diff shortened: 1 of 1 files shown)


docs/guides/exportOculusAPK.md
```diff
@@ -67,44 +67,46 @@ export PATH=$PATH:$ANDROID_HOME/platform-tools
...
 
-to install NDK directly run
+To install NDK directly, run:
 
 `sdkmanager "ndk-bundle"`
 
 ### Build APK
 
-Confirm that `$ANDROID_HOME` points to Android/Sdk directory
+Confirm that `$ANDROID_HOME` points to Android/SDK directory
```
chrisplatorres: `Android/Sdk` is actually the correct path here, not `SDK`

chrisplatorres:
Thanks, it looks good. The only change needed to merge is that there is one edit where it actually is editing an accurate path.

PR #1089, opened 2019-05-22
[Quest] Android XR Webview

modulesio:
This PR adds bindings for Android WebView to be used on Oculus mobile (Quest/Go).

The API should remain the same as it is on Desktop (which uses Chromium Embedded Framework). That is, we will continue to use 2D `<iframe>` composition on the `XRSession` as the end user DOM-to-texture entry point.

[WIP]


modulesio:
References:

https://stackoverflow.com/questions/44304105/unable-to-render-an-android-webview-into-an-external-texture-shared-between-c
https://stackoverflow.com/questions/17018960/create-a-webview-without-xml
http://www.felixjones.co.uk/neo%20website/Android_View/

chrisplatorres:
Would this apply to Magic Leap as well?

modulesio:
No, magic leap has no web view api in the regular sdk (it has in the lumin runtime engine).

But this web view jni binding should work with most android things, including gvr.

arpu:
there is now a magic leap polyfill for newer webxr api 
https://github.com/mrdoob/three.js/pull/16649   maybe this works too for the oculus browser

modulesio:
That’s a slightly different concern — this issue is about 2d planes rendering, not webxr.

modulesio:
This should be doing the thing now (injecting Oculus OS webviews into the XR scene per the iframe 2D API).

The major caveat that I haven't been able to work around is that if a webview is created while there is an active VR session, Oculus will decide to yank you out of VR entirely so that it can draw the page in Oculus browser (previous world is lost).

There's no _easy_ way around this since it looks like Oculus is enforcing this at the image level. We _could_ theoretically build our own chromium webview lib but this is nasty for several reasons, including space, maintenance burden, and many hours wasted on shepherding the compile. I don't think it's worth it for Studio cases.

The workaround is to construct 2D iframes before entering XR.

modulesio:
- [ ] hook up the usual browser integration events -- raycast to touch, click
- [ ] figure out a preload technique that allows 2d reality tabs to be loaded in the studio context

PR #1098, opened 2019-05-24
add FakeVRDisplay's gamepads from globalGamepads

chrisplatorres:
This PR adds and assigns FakeVRDisplay's gamepads from globalGamepads.

### DIFF
--- a/examples/realitytabs.html
+++ b/examples/realitytabs.html
@@ -1691,13 +1691,13 @@ window.addEventListener('click', () => {
 });
 window.addEventListener('mousedown', e => {
   if (window.document.pointerLockElement) {
-    var gamepads = navigator.getGamepads();
+    const gamepads = navigator.getGamepads();
     gamepads[1].buttons[1].pressed = true;
   }
 });
 window.addEventListener('mouseup', e => {
   if (window.document.pointerLockElement) {
-    var gamepads = navigator.getGamepads();
+    const gamepads = navigator.getGamepads();
     gamepads[1].buttons[1].pressed = false;
   }
 });
@@ -1750,7 +1750,7 @@ const intersectionSpecs = [
 ];
 window.addEventListener('keydown', e => {
   if (window.document.pointerLockElement) {
-    var gamepads = navigator.getGamepads();
+    const gamepads = navigator.getGamepads();
     switch (e.which) {
       case 87: { // W
         keys.up = true;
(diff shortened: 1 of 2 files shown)


chrisplatorres:
This is missing the cases other than `main` gamepads.

modulesio:
Should we be attaching these to the fake vr display going forward? This isn't a standard, so I think the more proper way would be to update the `realitytabs.html` example to use the proper APIs.

Also, I don't think we need to worry about non-main gamepads.

chrisplatorres:
I don't understand, is this not the standard?
`navigator.getGamepads();`
https://github.com/exokitxr/exokit/blob/master/examples/realitytabs.html#L1873

modulesio:
That part is fine; the part that's not is [decorating the display](https://github.com/exokitxr/exokit/pull/1098/files#diff-801f8b3e4342dd7a885d431f0dd9f72cR579).

Displays do not have gamepads, but you can query a matching gamepad with `navigator.getGamepads`.

chrisplatorres:
> I think the more proper way would be to update the realitytabs.html example to use the proper APIs.

> Displays do not have gamepads, but you can query a matching gamepad with navigator.getGamepads.
  

You are meaning to decouple the gamepads from the FakeVRDisplay?

modulesio:
> You are meaning to decouple the gamepads from the FakeVRDisplay?

Correct, that was never legal.

The `FakeVRDisplay` can and should still update the gamepads as required per emulation, but they should not be directly decorated on the display.

chrisplatorres:
Makes sense, should we merge this fix for now and decouple it in a PR from master?

modulesio:
No, I think that would be a regression. It seems the only reason for this code would be to fix `realitytabs.html`'s non-standard usage. That code is to be deleted with the studio anyway.

It's probably 20 lines in `realitytabs.html` that need to be changed to fix this usage.

chrisplatorres:
Ok, I am going to revert the current changes in this branch and update the PR to fix `realitytabs.html` usage.

modulesio:
Thanks, the fix would probably be basically the same thing, except in `realitytabs.html` instead of `VR.js`.

chrisplatorres:
Now working, not sure if all the `var gamepads = navigator.getGamepads();`'s is redundant.

modulesio:
Sweet! Should be ok, this code will be superceded by the studio.

PR #1102, opened 2019-05-25
Realitytabs.html  texture mipmapping 

chrisplatorres:
This increases the mipmap+filter quality for the `realitytabs.html` instructions texture.

Note: increasing Texture.minFilter to LinearMipMapLinearFilter results in:
```
THREE.WebGLRenderer: Texture is not power of two. Texture.minFilter should be se
t to THREE.NearestFilter or THREE.LinearFilter.
```

### DIFF
--- a/examples/realitytabs.html
+++ b/examples/realitytabs.html
@@ -928,11 +928,11 @@ const _makeRig = () => {
         THREE.UVMapping,
         THREE.ClampToEdgeWrapping,
         THREE.ClampToEdgeWrapping,
-        THREE.NearestFilter,
+        THREE.LinearMipMapLinearFilter,
         THREE.NearestFilter,
         THREE.RGBAFormat,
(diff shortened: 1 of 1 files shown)


modulesio:
Re: the latter, we can make the textures power-of-two to get mipmapping.

Or, can try a WebGL2 rendering context: https://discourse.threejs.org/t/warning-from-threejs-image-is-not-power-of-two/7085

chrisplatorres:
setting the renderer to webgl2 doesn't fix it, it just needs have `minFilter` not be `LinearMipMapLinearFilter` and instead be `LinearFilter`, this isn't an issue because the `maxFilter` can remain `LinearMipMapLinearFilter` with no issues.

The wanted quality bump is still there.

https://github.com/exokitxr/exokit/pull/1102

https://stackoverflow.com/a/29422240

PR #1104, opened 2019-05-25
Linear render loop

modulesio:
Exokit's reality layers has generally used a bunch of framebuffers (one or two per window, depending on AA mode) to effect composited rendering. This works great on desktop where memory bandwidth is huge, pipelines are short (GPU+CPU is very low latency), and allows for exokit to things like freeze framebuffers and reproject them itself.

However, this seems to not work great on mobile platforms which have deep tiled render pipelines, with also limited memory bandwidth.

The Oculus Quest in particular uses a triple-buffered framebuffer and generally encourages at least one frame of lag in the pipe, which is fixed by the space/time warp. With a multi-framebuffer blitting approach, this deep pipeline will be hitched on the constant blitting and synchronization of pixels that we are scribbling over.

The result is that the system runs at half or less GPU capacity -- for the reality tabs composition case -- since we cannot queue beyond the current frame (or even the current window).

With this PR, we do a linear walk over all of the windows in the graph each frame, with a fully shared but synchronized framebuffer direct to the eye. The result should be a perf boost on mobile systems.


modulesio:
Since we don't use the composition shaders in this PR, we can probably remove their creation + caching for a cleaner/faster initializing GL pipeline: https://github.com/exokitxr/exokit/blob/70eb33a89d8ccea2d1979cac7a6beddf9d8f8562/deps/exokit-bindings/windowsystem/src/windowsystem.cc#L115

modulesio:
This PR should give reality tabs 72 FPS no drops on Quest in my testing.

modulesio:
Before merging, we should add back an antialiasing pipeline. That is, MSAA textures swapchain that we blit prior to submit. This should be compatible with the render strategy.

@chrisplatorres Also mentioned that we can enable foveated rendering with this pipeline as well.

chrisplatorres:
Should foveated rendering be PRed into this branch?

modulesio:
Merged: https://github.com/exokitxr/exokit/pull/1105

modulesio:
There’s an a-frame + a-frame composition case to debug here.

When you render both, only the last layer shows. Looking into it.

modulesio:
Also, with this PR we actually do _not_ need any `session/display.layers`; the iframe's `.hidden` and entry into VR handles whether it will be layered.

I'm not sure if that's a good thing, but it might be the right default doing forward.

modulesio:
With Diego's PR, this should be getting into Moon Rider fine now.

Capture 26

modulesio:
In problem with this branch is that since it elides `glClear` for the purpose of not stomping on the main framebuffer, it will actually also elide `glClear` on other render targets like shadow maps.

You can see this on Supercraft lobby easily.

The way to solve this is to elide the `glClear` only for the main framebuffer. We already track it so that's not hard.

modulesio:
Here is an Oculus Quest performance profile for Supercraft with this branch:

Capture 27

modulesio:
The performance with Supercraft turned out to be the use of PCF soft shadow map shading (high texture usage).

modulesio:
The three things we need before merging this branch:

- [Fix session `.layers` not respected](https://github.com/exokitxr/exokit/pull/1133/files)
- [Re-add antialiasing](https://github.com/exokitxr/exokit/pull/1124)
- `realitytabs.html` hitches just by doing nothing in my testing -- not sure what that is

modulesio:
Continuing work in https://github.com/exokitxr/exokit/pull/1143.

This branch is still used by some folks so we won't delete it yet, but for the same reason, continuing in the other branch so it's less disruptive.

PR #1105, opened 2019-05-25
Check for and add foveated rendering for Oculus Mobile

chrisplatorres:
This PR checks for and adds foveated rendering for Oculus Mobile.


chrisplatorres:
Building the apk to test now.

chrisplatorres:
Tested at [highest FFR setting (3)](https://developer.oculus.com/documentation/mobilesdk/latest/concepts/mobile-ffr/), no notable regression when loaded into a-painter with exo-bots in corners of screen.

modulesio:
Thanks! At some point I think we'll want to expose this to WebVR/WebXR, but I think the smartest default is "on".

PR #1109, opened 2019-05-26
Check for custom exokit:// urls

chrisplatorres:
This PR proposes adding to the current window scripts a check for the custom `exokit://` URLs.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1059,16 +1059,7 @@ const _start = () => {
     if (u === '.') {
       console.warn('NOTE: You ran `exokit . <url>`\n(Did you mean to run `node . <url>` or `exokit <url>` instead?)')
     }
-    if (u.includes("exokit://")){
-      u = u.replace(/^exokit:\/\//, '');
-    }
-    if (u.includes("https//") || u.includes("http//")){
-      u = u.replace(/(^\w+|^)\/\//, '');
-    }
-    if (u.includes("https://") || u.includes("http://")){
-      u = u.replace(/(^\w+:|^)\/\//, '');
-    }
-    u = 'http://' + u;
+    u = u.replace(/^exokit:/, '');
     if (args.tab) {
       u = u.replace(/\/?$/, '/');
       u = `${realityTabsUrl}?t=${encodeURIComponent(u)}`
--- a/src/index.js
+++ b/src/index.js
@@ -1059,7 +1059,7 @@ const _start = () => {
     if (u === '.') {
       console.warn('NOTE: You ran `exokit . <url>`\n(Did you mean to run `node . <url>` or `exokit <url>` instead?)')
     }
-    u = u.replace(/^exokit:/, '');
+    u = u.replace(/^exokit:\/\//, '');
     if (args.tab) {
       u = u.replace(/\/?$/, '/');
       u = `${realityTabsUrl}?t=${encodeURIComponent(u)}`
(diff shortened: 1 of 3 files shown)


modulesio:
If the intent is for Exokit to support the `exokit://` URL scheme, can we just do that in `index.js`?

Or is there a good reason the windows start shell script needs to be involved here?

chrisplatorres:
This was put together in a very short-sighted/"don't worry be hacky" way, will start over and do it in `index.js` instead.

src/index.js
```diff
@@ -1059,7 +1059,16 @@ const _start = () => {
...
+    if (u.includes("exokit://")){
+      u = u.replace(/^exokit:\/\//, '');
+    }
+    if (u.includes("https//") || u.includes("http//")){
+      u = u.replace(/(^\w+|^)\/\//, '');
+    }
+    if (u.includes("https://") || u.includes("http://")){
+      u = u.replace(/(^\w+:|^)\/\//, '');
+    }
+    u = 'http://' + u;
```
chrisplatorres: Should this be http:// or https:// by default?

chrisplatorres:
Tested cases:

`exokit://https://aframe.io/a-painter`
`exokit://https//aframe.io/a-painter`
`https://aframe.io/a-painter`
`https//aframe.io/a-painter`
`exokit://aframe.io/a-painter`
`exokit://http://aframe.io/a-painter`
`exokit://http//aframe.io/a-painter`
`http://aframe.io/a-painter`
`http//aframe.io/a-painter`

src/index.js
```diff
@@ -1059,7 +1059,16 @@ const _start = () => {
     if (u === '.') {
       console.warn('NOTE: You ran `exokit . <url>`\n(Did you mean to run `node . <url>` or `exokit <url>` instead?)')
     }
-    u = u.replace(/^exokit:/, '');
+    if (u.includes("exokit://")){
```
modulesio: Probably should be regexes. It's not unthinkable that a URL would be in a query string, which would throw all of this off.

src/index.js
modulesio: `http://`

src/index.js
```diff
@@ -1059,7 +1059,16 @@ const _start = () => {
...
+      u = u.replace(/^exokit:\/\//, '');
+    }
+    if (u.includes("https//") || u.includes("http//")){
```
modulesio: This would not be valid so I don't see the need to check it.

chrisplatorres: It is a case in discord when using `<exokit://https://test.com>` it turns it into `<exokit://https//test.com`.

src/index.js
chrisplatorres: > It's not unthinkable that a URL would be in a query string

What would that look like? `someurl.com?h=exokit://othersite.com`

modulesio: More likely it would be `http://`, but yes.

src/index.js
modulesio: If it's a bug in discord, we should fix it there rather than adding magic to the engine. Is it possible to fix on the Discord end?

chrisplatorres: Pending response from Discord.

modulesio:
Actually -- shouldn't it be `exokit:` only? For protocol handlers I think that's how it's done. The `//` would be the start of the legitimate path.

So a correct exokit link would be `exokit:https://aframe.io/a-painter`

chrisplatorres:
Both seem to work.

PR #1112, opened 2019-05-26
Web Workers importScripts

modulesio:
Web Worker `importScripts`, which allows one to perform synchronous injection of code from the network, is a pretty weird feature, since it's the only case (other than deprecated `XMLHttpRequest` sync mode) where you can write code that blocks on the network.

The way we do this in Exokit is create a sub-worker to a `request.js` script that blocks on atomics to do the actual request asynchronously in a separate thread. This works with no native modules, only regular Javascript + node.js features.

The bug here was that `request.js` was missing. This PR implements it and patches in the worker environment calls to capture the requests.

### DIFF
--- a/src/WindowBase.js
+++ b/src/WindowBase.js
@@ -69,6 +69,7 @@ const _normalizeUrl = src => {
     return src;
   }
 };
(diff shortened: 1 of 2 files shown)


modulesio:
Note that there is currently a `5MB` limit for this request pipeline.

We could increase it by doing a loop dance with atomics to transfer over the buffer gradually, but this PR does not do that yet. Will comment to that effect.

PR #1114, opened 2019-05-26
HTMLAudioElement/HTMLVideoElement loaded events

modulesio:
There are `loadeddata` and `loadedmetadata` for media elements that we forgot to emit when we finish loading media. This PR adds them.

Moon Rider was depending on these.


modulesio:
MDN links:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/loadeddata_event
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/loadedmetadata_event

Issue #1078, opened 2019-05-21
Load Moon Rider in Exokit


chrisplatorres:
Windows 10
Node v12.2.0
`user-parallel` branch of Exokit
site url: https://moonrider.xyz/

command ran: `$ node --experimental-worker . -h -t https://moonrider.xyz/`

[View full log in gist](https://gist.github.com/chrisplatorres/30fb09b91595674740a0e4b5050739b7)

modulesio:
Thanks! I already see a few things like Worker `xmlhttprequest` require misnaming.

modulesio:
Looks like it's mostly getting stuck in the preload image phase.

chrisplatorres:
Thanks, the [require renaming](https://github.com/exokitxr/exokit/commit/63ef2340af3f5f33bdc332e340b9ea7103609315) fixed some of the errors, I notice now the moon rider audio is loading for a few seconds.

> Looks like it's mostly getting stuck in the preload image phase.
  
`[gpu-preloader] Error preloading, image not loaded s {`


Yes, where would I be looking for the preloading stage to be getting stuck in the engine?
  
----------------------------------------------------

```
TypeError: Cannot read property '1' of undefined
    at i.getDistance (build/build.js:180:68102)
    at i.tick (build/build.js:180:67699)
    at HTMLElement.value (vendor/aframe-master.min.js:255:11226)
    at HTMLElement.value (vendor/aframe-master.min.js:255:11760)
    at vendor/aframe-master.min.js:386:132
    at m (vendor/aframe-master.min.js:89:21552)
    at t (vendor/aframe-master.min.js:87:3246)
    at _renderLocal (C:\Users\chris\github\exokit\src\Window.js:1305:15)
    at window.tickAnimationFrame (C:\Users\chris\github\exokit\src\Window.js:1364:5)
    at global.onrunasync.method (C:\Users\chris\github\exokit\src\Window.js:1529:19)
```
https://github.com/exokitxr/exokit/blob/user-parallel/src/Window.js#L1529

```
TypeError: Cannot read property '0' of undefined
    at r.decompose (vendor/aframe-master.min.js:92:27113)
    at i.<anonymous> (vendor/aframe-master.min.js:170:4261)
    at i.tick (vendor/aframe-master.min.js:170:1525)
    at HTMLElement.value (vendor/aframe-master.min.js:255:11226)
    at HTMLElement.value (vendor/aframe-master.min.js:255:11760)
[...]

PR #1116, opened 2019-05-27
XMLHttpRequest Blob support

modulesio:
XMLHttpRequest did not support support `blob://` URLs, and did not output proper `Blob` when `requestType === 'blob'`. This PR adds that support.

Note that `fetch` has supported for this for a long time, so we simply re-use that `Blob` implementation.

Ingests https://github.com/modulesio/window-xhr/pull/4.
Needed for https://github.com/exokitxr/exokit/issues/1078.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "window-fetch": "0.0.10",
     "window-ls": "0.0.1",
     "window-selector": "0.0.6",
-    "window-xhr": "0.0.29",
+    "window-xhr": "0.0.30",
     "ws": "^6.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


modulesio:
Note: this PR also removes the unused exokit `download` flag, which was entangling the logic. This also shaves some functions from the network calls.

modulesio:
Since Blobs can, and are, created in `Worker`/different context from the `window` that uses them, we'll need a messaging solution to handle this buffering.

Currently thinking a native cache, which would also help alleviate the `5MB` importScripts/synchronous `request.js` limit, as well as not over-commit the memory for it...

Issue #981, opened 2019-04-25
Oculus Mobile Multiview rendering


modulesio:
The [OVR_multiview2 extension](https://www.khronos.org/registry/OpenGL/extensions/OVR/OVR_multiview2.txt) allows for a cleaner GL call path (if the client uses it).

[THREE.js is adding support]https://github.com/mrdoob/three.js/pull/16316/files/08ded2c0b55919b02d691dd0fa3b8806123bc5f3..1e13fbdd086cc1eaf9cda0994b29d0335480ae4c) and we might as well follow suit.

This should be relatively straightforward since we already just bind directly to OpenGL.

chrisplatorres:
Relevant link https://developer.oculus.com/documentation/mobilesdk/latest/concepts/mobile-multiview/

chrisplatorres:
I built a multiview three.js here: https://chrisplatorres.github.io/three.js/build/three.js

just add it and:
```js
const renderer = new THREE.WebGLRenderer( { antialias: true, multiview: true } );
```

chrisplatorres:
Closed by https://github.com/exokitxr/exokit/pull/1122.

PR #1122, opened 2019-05-27
Add OVR_multiview2 and OVR_multiview_multisampled_render_to_texture extensions 

chrisplatorres:
from #981 this PR adds `OVR_multiview2` and `OVR_multiview_multisampled_render_to_texture extensions`.

-------------------------------------------
  
### Notes: 
Multiview is mainly working on the CPU by [reducing the number of draw calls](https://community.arm.com/developer/tools-software/graphics/b/blog/posts/optimizing-virtual-reality-understanding-multiview) you need to issue in order to draw your scene.
  
CPU-bound experiences will benefit from multi-view. Typically a [CPU usage reduction of 25% - 50% is possible.](https://developer.oculus.com/documentation/oculus-browser/latest/concepts/browser-multiview/)


chrisplatorres:
I see multiview in the WebXR [XR.js](https://github.com/exokitxr/exokit/blob/6977b177b6d848acc91205788a09ad668d817a30/src/XR.js#L357), should there be multiview attribute in [WebVR](https://github.com/mrdoob/three.js/pull/15875/files?file-filters%5B%5D=.js#diff-fdee169d721b24ed863f98ee8b9f67f4R56)'s VR.js?

chrisplatorres:
Still need to setup on Oculus side.

modulesio:
> I see multiview in the WebXR XR.js, should there be multiview attribute in WebVR's VR.js?

That depends on what the spec says on the matter. I think the answer is no.

chrisplatorres:
> WebVR 1.1 has also been extended to provide opaque multiview framebuffer support, with corresponding views / viewports. The upcoming WebXR standard already has multiview support.

https://developer.oculus.com/documentation/oculus-browser/latest/concepts/browser-multiview/#webvr-extension

modulesio:
☝️ In that case we should follow suit in our `VR.js`, per spec.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4854,6 +4854,8 @@ const char *webglExtensions[] = {
   "OES_texture_half_float",
   "OES_texture_half_float_linear",
   "OES_vertex_array_object",
+  "OVR_multiview",
```
chrisplatorres: This should be `multiview2`

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4919,6 +4952,8 @@ const char *webglExtensions[] = {
   "OES_texture_half_float",
   "OES_texture_half_float_linear",
   "OES_vertex_array_object",
+  "OVR_multiview2",
```
modulesio: We probably do not want to advertise this if it does not exist -- it might not on desktops.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -626,6 +626,12 @@ void setGlConstants(T &proto) {
   JS_GL_SET_CONSTANT("PIXEL_PACK_BUFFER_BINDING" , 0x88ED);
   JS_GL_SET_CONSTANT("PIXEL_UNPACK_BUFFER_BINDING", 0x88EF);
 
+  // OVR_multiview2 in draft
+  JS_GL_SET_CONSTANT("FRAMEBUFFER_ATTACHMENT_TEXTURE_NUM_VIEWS_OVR", 0x9630);
```
modulesio: Does this need to be global? I'm not sure if the spec says it is, or if it's only on the extension.

chrisplatorres: Probably not, it should be through OVR_multiview2, will fix.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -2481,6 +2491,29 @@ NAN_METHOD(WebGLRenderingContext::SetTopLevel) {
...
+  GLint baseViewIndex = TO_INT32(info[4]);
+  GLsizei numViews = TO_UINT32(info[5]);
+
+  glframebufferTextureMultiviewOVR(target, attachment, texture, level, baseViewIndex, numViews);
+}
+
+NAN_METHOD(WebGLRenderingContext::FramebufferTextureMultisampleMultiviewOVR) {
+  GLenum target = TO_UINT32(info[0]);
+  GLenum attachment = TO_UINT32(info[1]);
+  GLuint texture = TO_INT32(info[2]);
```
modulesio: This one will need to be extracted out of `.id`.

Our implementation of `WebGLTexture` is actually `{id: GLID}` (in most implementations it is completely opaque). But it is not a `GLuint`.

chrisplatorres: I made it as `GLuint` based off [here](https://github.com/KhronosGroup/OpenGL-Registry/blob/master/extensions/OVR/OVR_multiview_multisampled_render_to_texture.txt#L53).

modulesio: It's correct that it's a `GLuint` at the OpenGL level. But WebGL does not use ints for textures. It uses `WebGLTexture`, which contains the int in an abstract way. In our case that's concretely `tex.id`.

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: 8b2c577
and
8cc91fd

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: 0a608de
and
0cbd3f1

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: 8ece516

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -2481,6 +2485,29 @@ NAN_METHOD(WebGLRenderingContext::SetTopLevel) {
   gl->topLevel = topLevel;
 }
 
+NAN_METHOD(WebGLRenderingContext::FramebufferTextureMultiviewOVR) {
+  GLenum target = TO_UINT32(info[0]);
+  GLenum attachment = TO_UINT32(info[1]);
+  WebGLTexture? texture = TO_INT32(info[2]);
```
modulesio: This will not build...

chrisplatorres: 114b64b

deps/exokit-bindings/webglcontext/src/webgl.cc
modulesio: The problem is, if we don't advertise it for the case where it does exist, then user code will think it's not supported.

More generally, I think we should not hardcode this extension list, but check each one with the hardware and return the whitelisted WebGL ones.

chrisplatorres:
current build error:
```
../deps/exokit-bindings/webglcontext/src/webgl.cc:1162:70: error: 'FramebufferTextureMultiviewOVR' was not declared in this scope
   Nan::SetMethod(proto, "framebufferTextureMultiviewOVR", glCallWrap<FramebufferTextureMultiviewOVR>);
                                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../deps/exokit-bindings/webglcontext/src/webgl.cc:1162:101: error: no matching function for call to 'SetMethod(v8::Local<v8::ObjectTemplate>&, const char [31], <unresolved overloaded function type>)'
   Nan::SetMethod(proto, "framebufferTextureMultiviewOVR", glCallWrap<FramebufferTextureMultiviewOVR>);
                                                                                                     ^
In file included from /app/deps/exokit-bindings/webglcontext/include/webgl.h:11:0,
                 from ../deps/exokit-bindings/webglcontext/src/webgl.cc:4:
/app/node_modules/nan/nan.h:2474:13: note: candidate: template<class T, template<class> class HandleType> void Nan::SetMethod(HandleType<T>, const char*, Nan::FunctionCallback)
 inline void SetMethod(
             ^~~~~~~~~
/app/node_modules/nan/nan.h:2474:13: note:   template argument deduction/substitution failed:
../deps/exokit-bindings/webglcontext/src/webgl.cc:1162:59: note:   cannot convert 'glCallWrap<<expression error> >' (type '<unresolved overloaded function type>') to type 'Nan::FunctionCallback {aka void (*)(const Nan::FunctionCallbackInfo<v8::Value>&)}'
   Nan::SetMethod(proto, "framebufferTextureMultiviewOVR", glCallWrap<FramebufferTextureMultiviewOVR>);
                                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../deps/exokit-bindings/webglcontext/src/webgl.cc:1163:81: error: 'FramebufferTextureMultisampleMultiviewOVR' was not declared in this scope
   Nan::SetMethod(proto, "FramebufferTextureMultisampleMultiviewOVR", glCallWrap<FramebufferTextureMultisampleMultiviewOVR>);
                                                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../deps/exokit-bindings/webglcontext/src/webgl.cc:1163:123: error: no matching function for call to 'SetMethod(v8::Local<v8::ObjectTemplate>&, const char [42], <unresolved overloaded function type>)'
   Nan::SetMethod(proto, "FramebufferTextureMultisampleMultiviewOVR", glCallWrap<FramebufferTextureMultisampleMultiviewOVR>);
                                                                                                                           ^
In file included from /app/deps/exokit-bindings/webglcontext/include/webgl.h:11:0,
                 from ../deps/exokit-bindings/webglcontext/src/webgl.cc:4:
/app/node_modules/nan/nan.h:2474:13: note: candidate: template<class T, template<class> class HandleType> void Nan::SetMethod(HandleType<T>, const char*, Nan::FunctionCallback)
 inline void SetMethod(
             ^~~~~~~~~
/app/node_modules/nan/nan.h:2474:13: note:   template argument deduction/substitution failed:
../deps/exokit-bindings/webglcontext/src/webgl.cc:1163:70: note:   cannot convert 'glCallWrap<<expression error> >' (type '<unresolved overloaded function type>') to type 'Nan::FunctionCallback {aka void (*)(const Nan::FunctionCallbackInfo<v8::Value>&)}'
   Nan::SetMethod(proto, "FramebufferTextureMultisampleMultiviewOVR", glCallWrap<FramebufferTextureMultisampleMultiviewOVR>);
                                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: Something like: iterate through the current list and return the whitelisted result here: https://github.com/exokitxr/exokit/blob/add-multiview/deps/exokit-bindings/webglcontext/src/webgl.cc#L4968

chrisplatorres: 46be25f

chrisplatorres:
this is failing on android/Oculus Mobile's travis, but passing on linux/macOS:

```
  ../../../../../../build/libexokit/libexokit.a(webgl.o): In function `WebGLRenderingContext::FramebufferTextureMultiviewOVR(Nan::FunctionCallbackInfo<v8::Value> const&)':
  webgl.cc:(.text+0x1027c): undefined reference to `glFramebufferTextureMultiviewOVR'
  ../../../../../../build/libexokit/libexokit.a(webgl.o): In function `WebGLRenderingContext::FramebufferTextureMultisampleMultiviewOVR(Nan::FunctionCallbackInfo<v8::Value> const&)':
  webgl.cc:(.text+0x10674): undefined reference to `glFramebufferTextureMultisampleMultiviewOVR'
```

modulesio:
From Discord:

```
ack glFramebufferTextureMultisampleMultiviewOVR

ndk-bundle/toolchain/sysroot/usr/include/GLES2/gl2ext.h
3440:GL_APICALL void GL_APIENTRY glFramebufferTextureMultisampleMultiviewOVR (GLenum target, GLenum attachment, GLuint texture, GLint level, GLsizei samples, GLint baseViewIndex, GLsizei numViews);
```

If this is indeed the case, we probably need to include `GLES2/gl2ext.h` in `webgl.h`, for `ANDROID`.

chrisplatorres:
All checks are passing, and tested/confirmed on Oculus Quest and Windows 10.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4940,10 +4998,18 @@ NAN_METHOD(WebGLRenderingContext::GetSupportedExtensions) {
...
-    // char *extension = (char *)glGetStringi(GL_EXTENSIONS, i);
-    result->Set(i, JS_STR(webglExtensions[i]));
+    char *extension = (char *)glGetStringi(GL_EXTENSIONS, i);
+    if(extension == webglExtensions[i]){
+      result->Set(i, JS_STR(webglExtensions[i]));
+    }
+  }
+  for (GLint i = 0; i < numExtensions; i++) {
+    string currentExtension = string(webglExtensions[i]);
+    bool hasWebGL = currentExtension.find("WEBGL");
```
modulesio: I think there are corner cases in this code. There's no guarantee that any extension called `WEBGL` will not have been already included -- in which case our list would double up.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4940,10 +4998,18 @@ NAN_METHOD(WebGLRenderingContext::GetSupportedExtensions) {
...
+      if(hasWebGL) {
+        result->Set(i, JS_STR(webglExtensions[i]));
```
modulesio: This will overwrite previous entries.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4940,10 +4998,18 @@ NAN_METHOD(WebGLRenderingContext::GetSupportedExtensions) {
   int numExtensions = sizeof(webglExtensions)/sizeof(webglExtensions[0]);
   Local<Array> result = Nan::New<Array>(numExtensions);
   for (GLint i = 0; i < numExtensions; i++) {
-    // char *extension = (char *)glGetStringi(GL_EXTENSIONS, i);
-    result->Set(i, JS_STR(webglExtensions[i]));
+    char *extension = (char *)glGetStringi(GL_EXTENSIONS, i);
+    if(extension == webglExtensions[i]){
+      result->Set(i, JS_STR(webglExtensions[i]));
```
modulesio: This will leave holes in the output array.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -930,6 +935,21 @@ ColorMaskState &ColorMaskState::operator=(const ColorMaskState &colorMaskState)
 std::pair<Local<Object>, Local<FunctionTemplate>> WebGLRenderingContext::Initialize(Isolate *isolate) {
   // Nan::EscapableHandleScope scope;
 
+#if defined(ANDROID) || defined(LUMIN)
```
modulesio: Is this the right place to do the initialization? I think it will happen once for every `window`. We should endeavor to do this initialization only once, globally.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -930,6 +935,21 @@ ColorMaskState &ColorMaskState::operator=(const ColorMaskState &colorMaskState)
...
+  glFramebufferTextureMultiviewOVRExt = (PFNGLFRAMEBUFFERTEXTUREMULTIVIEWOVR)eglGetProcAddress("glFramebufferTextureMultiviewOVR");
+  if (!glFramebufferTextureMultiviewOVRExt) {
+      std::cerr << "Can not get proc address for glFramebufferTextureMultiviewOVR." << std::endl;
+      sleep(1);
```
modulesio: There is probably no need to sleep, or perhaps even abort. I think as long as we initialize to `nullptr` and log the warning, it would be fine to continue -- only applications that use multiview on ANDROID would be affected by it being missing.

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: 8a5cd49

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: 8a5cd49

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: 93391d1

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -5000,14 +5000,17 @@ NAN_METHOD(WebGLRenderingContext::GetSupportedExtensions) {
   for (GLint i = 0; i < numExtensions; i++) {
     char *extension = (char *)glGetStringi(GL_EXTENSIONS, i);
     if(extension == webglExtensions[i]){
-      result->Set(i, JS_STR(webglExtensions[i]));
+      int numResults = result->Length();
```
chrisplatorres: it may be redundant to add this assign this to a variable, instead just use `result->Length()+1` instead of `numResults`

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: 5bf4b3f

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: 8ae3505

chrisplatorres:
Added fixes per requests, and confirmed working during test:

```
$ node . https://chrisplatorres.github.io/three.js/examples/webgl_multiple_views_multiview.html
THREE.WebGLRenderer 105
WebGLMultiview {
  attachRenderTarget: [Function: attachRenderTarget],
  detachRenderTarget: [Function: detachRenderTarget],
  isAvailable: [Function: isAvailable],
  isEnabled: [Function: isEnabled],
  getNumViews: [Function: getNumViews],
  updateCameraProjectionMatricesUniform: [Function: updateCameraProjectionMatric
esUniform],
  updateCameraViewMatricesUniform: [Function: updateCameraViewMatricesUniform],
  updateObjectMatricesUniforms: [Function: updateObjectMatricesUniforms]
}
```

deps/exokit-bindings/webglcontext/src/webgl.cc
modulesio: I think there will still be holes at the end of the array, since the [length is preallocated](https://github.com/exokitxr/exokit/commit/8a5cd4929f9fd348b3cdc0fa019dd6ae323e946d#diff-9830800c0783d1a603a0e5fc068f7d67R4999).

deps/exokit-bindings/webglcontext/src/webgl.cc
modulesio: No worries, this would get optimized away.

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: e760558

chrisplatorres:
Confirmed working on Oculus Mobile and Desktop/win10.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4938,12 +4997,26 @@ NAN_METHOD(WebGLRenderingContext::GetSupportedExtensions) {
   // glGetIntegerv(GL_NUM_EXTENSIONS, &numExtensions);
 
   int numExtensions = sizeof(webglExtensions)/sizeof(webglExtensions[0]);
-  Local<Array> result = Nan::New<Array>(numExtensions);
+  Local<Array> result = Nan::New<Array>(0);
   for (GLint i = 0; i < numExtensions; i++) {
-    // char *extension = (char *)glGetStringi(GL_EXTENSIONS, i);
-    result->Set(i, JS_STR(webglExtensions[i]));
+    char *extension = (char *)glGetStringi(GL_EXTENSIONS, i);
+    if(extension == webglExtensions[i]){
```
modulesio: I don't think this is how string comparison works in C++. You cannot just compare `char *` pointers -- I don't think this will ever be true.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4938,12 +4997,26 @@ NAN_METHOD(WebGLRenderingContext::GetSupportedExtensions) {
...
+      string currentExtension = string(webglExtensions[i]);
+      bool hasWebGL = currentExtension.find("WEBGL");
```
modulesio: This is probably not a reliable check -- we should instead manually chop the array above to make our intent clear. It would also be faster.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4938,12 +4997,26 @@ NAN_METHOD(WebGLRenderingContext::GetSupportedExtensions) {
...
+        if(!hasWebGL) {
+          int numResults = result->Length();
+          result->Set(numResults+1, JS_STR(webglExtensions[i]));
```
modulesio: Won't this leave holes every other entry?

chrisplatorres:
site to test `getSupportedExtensions`:

```html
<!DOCTYPE html>
<html>
<canvas id="canvas"></canvas>
<script>
var canvas = document.getElementById('canvas');
gl = canvas.getContext('webgl');

var extensions = gl.getSupportedExtensions();
console.log("extensions: " + extensions);
</script>
```

chrisplatorres:
Confirmed accurate results on Desktop/windows 10 with the code from  https://github.com/exokitxr/exokit/pull/1122#issuecomment-498462363:

```
$ node . testext.html
extensions: EXT_blend_minmax,EXT_texture_filter_anisotropic,WEBGL_color_buffer_float,WEBGL_compressed_texture_astc,WEB
GL_compressed_texture_atc,WEBGL_compressed_texture_etc,WEBGL_compressed_texture_etc1,WEBGL_compressed_texture_pvrtc,WE
BGL_compressed_texture_s3tc,WEBGL_compressed_texture_s3tc_srgb,WEBGL_debug_renderer_info,WEBGL_debug_shaders,WEBGL_dep
th_texture,WEBGL_draw_buffers,WEBGL_lose_context
```

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4938,12 +4997,33 @@ NAN_METHOD(WebGLRenderingContext::GetSupportedExtensions) {
...
+  Local<Array> result = Nan::New<Array>(0);
+
+  GLint n=0;
+  glGetIntegerv(GL_NUM_EXTENSIONS, &n);
+
+  for (GLint i=0; i<n; i++)
+  {
+    const char* extension = (const char*)glGetStringi(GL_EXTENSIONS, i);
+
+    for (GLint j = 0; j < numExtensions; j++) {
```
modulesio: This shouldn't be `GLint` if it's not passed to OpenGL.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4938,12 +4997,33 @@ NAN_METHOD(WebGLRenderingContext::GetSupportedExtensions) {
...
+      bool matchesSupportedExtension = (strcmp(extension+3,webglExtensions[j]) == 0);
```
modulesio: What is `+3`? Is that guaranteed to never overflow?

chrisplatorres: `+3` is to compare against `weblExtensions` (which do not lead with `GL_`) against the GL_EXTENSIONS (which have `GL_` as the first 3 characters). See example output here: https://www.geeks3d.com/20100722/tips-how-to-get-the-list-of-the-opengl-extensions-with-a-core-profile/

gloutputCapture

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: 232be5b

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: There aren't any obvious cases where it would overflow, based on the values it outputs. (`GL_xxx`)

chrisplatorres: It looks like there is also `WGL_`, so a regex that filters for characters up to first `_` might be better.

modulesio: The problem is that there is no guarantee this code won't explode if some driver decides a different format for an extension string. Can this code be written without making any assumptions about the GL implementation?

chrisplatorres: The first thing that comes to mind is to see if characters of `webglExtension[i]` are found in `extension`, but that would be assuming things as well.

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4938,12 +4997,33 @@ NAN_METHOD(WebGLRenderingContext::GetSupportedExtensions) {
...
+            int numResults = result->Length();
+            result->Set(numResults, JS_STR(webglExtensions[j]));
+          }
+      }
+    }
+  }
   for (GLint i = 0; i < numExtensions; i++) {
-    // char *extension = (char *)glGetStringi(GL_EXTENSIONS, i);
-    result->Set(i, JS_STR(webglExtensions[i]));
+    bool hasWebGL = (strncmp(webglExtensions[i], "WEBGL", 5) == 0);
```
modulesio: If the extension was already included in the above list, it would be duplicated here, no?

modulesio: Actually, never mind, the above list generation checks for non-webgl. 👍

deps/exokit-bindings/webglcontext/src/webgl.cc
```diff
@@ -4938,12 +4997,33 @@ NAN_METHOD(WebGLRenderingContext::GetSupportedExtensions) {
...
+  glGetIntegerv(GL_NUM_EXTENSIONS, &n);
+
+  for (GLint i=0; i<n; i++)
+  {
+    const char* extension = (const char*)glGetStringi(GL_EXTENSIONS, i);
+
+    for (GLint j = 0; j < numExtensions; j++) {
+      bool matchesSupportedExtension = (strcmp(extension+3,webglExtensions[j]) == 0);
+      if(matchesSupportedExtension){
+        bool hasWebGL = (strncmp(webglExtensions[j], "WEBGL", 5) == 0);
```
modulesio: Can we manually split the list above into WEBGL and non-WEBGL? This seems a static segregation we don't need to be testing at runtime.

deps/exokit-bindings/webglcontext/src/webgl.cc
modulesio: What about if we used `std::string` and did a safe substring? That probably wouldn't crash regardless of what the driver returns.

chrisplatorres: The list from https://www.geeks3d.com/20100722/tips-how-to-get-the-list-of-the-opengl-extensions-with-a-core-profile/  has multiple substrings, would cases like that be relevant here?:
```
GL extension 191: GL_NV_texture_shader
GL extension 192: GL_NV_texture_shader2
GL extension 193: GL_NV_texture_shader3
GL extension 194: GL_NV_transform_feedback
GL extension 195: GL_NV_transform_feedback2
GL extension 196: GL_NV_vertex_array_range
GL extension 197: GL_NV_vertex_array_range2
GL extension 200: GL_NV_vertex_program
GL extension 201: GL_NV_vertex_program1_1
GL extension 202: GL_NV_vertex_program2
GL extension 203: GL_NV_vertex_program2_option
GL extension 204: GL_NV_vertex_program3
```

deps/exokit-bindings/webglcontext/src/webgl.cc
chrisplatorres: 6e5d4f6

deps/exokit-bindings/webglcontext/src/webgl.cc
modulesio: Yes, quite possibly.

Though I would expect the webgl extensions to be listed by the driver under a non-prefixed form, which would be the one we need to check.

chrisplatorres: note from discord:
>  Can probably just hardcode on compile for multiview extension support in the list / hardcode as available in all cases, since that seems to be the case

chrisplatorres: Based on the note^ 436dba4

modulesio:
Thanks! 👏

PR #1124, opened 2019-05-27
Linear render loop MSAA

modulesio:
The [linear walk render loop](https://github.com/exokitxr/exokit/pull/1104) for reality tabs has a different, minimal set of framebuffers optimized for a deep mobile rendering pipeline.

However, we lost antialiasing support: we have a single framebuffer in the chain, which is not multisampled.

This PR adds back a multisampled framebuffer path that works almost the same as the non-multisampled case, except with a final downsample before submit. Other than that, the same swapping discipline is followed, with triple-buffering of the multisampled framebuffers for Oculus.

### DIFF
--- a/deps/exokit-bindings/egl/src/egl.cc
+++ b/deps/exokit-bindings/egl/src/egl.cc
@@ -32,7 +32,31 @@ void Uninitialize() {
   eglTerminate(display);
 }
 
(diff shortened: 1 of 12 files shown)


modulesio:
This should have working AA on all platforms, including Oculus Quest.

PR #1127, opened 2019-05-28
Bugfix VRDisplay near/far clip planes editability

modulesio:
`VRDisplay.depth{Near,Far}` was exposed on `FakeVRDisplay` but not the parent `VRDisplay`. Therefore move it to the parent display.

This is used by some legacy WebVR experiences, which will look wrong without it.

### DIFF
--- a/src/VR.js
+++ b/src/VR.js
@@ -270,6 +270,19 @@ class VRDisplay extends EventEmitter {
     };
   }
 
+  get depthNear() {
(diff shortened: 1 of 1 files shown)


modulesio:
We might need to patch this for the Oculus Mobile case as well: https://github.com/exokitxr/exokit/blob/3a5c85be5b3924956e3cce84263336dab22a7c7e/deps/oculus-mobile/src/oculus-context.cpp#L300

PR #1134, opened 2019-05-29
Add reconfigurable launch arguments in main.cpp

modulesio:
This allows one to set a different site + arguments that Exokit will use on startup, in `main.cpp`.

This is useful when creating standalone apps where you cannot pass command line arguments manually.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -408,7 +408,7 @@ int main(int argc, char **argv) {
 
         offset += strlen(srcArgString) + 1;
       } else {
-        LOG_ERROR("node command line arguments overflow");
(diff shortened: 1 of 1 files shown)


chrisplatorres:
How would this be specified for a standalone app? Perhaps this section of the docs should be updated https://github.com/exokitxr/exokit/blob/master/docs/guides/exportOculusAPK.md#set-default-site

modulesio:
It's probably not time to promote this yet. This was an internal use case, but we needed something cleaner than the previous mess.

I think a better solution would be to go through a more friendly stack like `studio`, than documenting internal hacks.

PR #1141, opened 2019-05-29
Bugfix haptic pulse support

modulesio:
Exokit's OpenVR binding has supported haptics since forever, but it regressed in window parallelization, where the data flow for VR contexts is much more isolated.

This reroutes the haptic pulse activations to reach `index.js`, handles them there, and opens up the code path for other VR system bindings to handle haptics there too.


modulesio:
This PR also uses `console.warn` in "fake VR" mode so you can at least see the haptics when developing even if you can't feel them.

PR #1143, opened 2019-05-29
Linear render loop

modulesio:
Exokit's reality layers has generally used a bunch of framebuffers (one or two per window, depending on AA mode) to effect composited rendering. This works great on desktop where memory bandwidth is huge, pipelines are short (GPU+CPU is very low latency), and allows for exokit to things like freeze framebuffers and reproject them itself.

However, this seems to not work great on mobile platforms which have deep tiled render pipelines, with also limited memory bandwidth.

The Oculus Quest in particular uses a triple-buffered framebuffer and generally encourages at least one frame of lag in the pipe, which is fixed by the space/time warp. With a multi-framebuffer blitting approach, this deep pipeline will be hitched on the constant blitting and synchronization of pixels that we are scribbling over.

The result is that the system runs at half or less GPU capacity -- for the reality tabs composition case -- since we cannot queue beyond the current frame (or even the current window).

With this PR, we do a linear walk over all of the windows in the graph each frame, with a fully shared but synchronized framebuffer direct to the eye. The result should be a perf boost on mobile systems.

(c.f. https://github.com/exokitxr/exokit/pull/1104)


modulesio:
Now supports MSAA (c.f. https://github.com/exokitxr/exokit/pull/1124).

Issue #747, opened 2019-01-18
Update WebXR API


modulesio:
Per chrome+three.js changes: https://github.com/mrdoob/three.js/pull/15604

PR #1151, opened 2019-05-30
WebXR API update

modulesio:
Goal: make the latest versions of https://immersive-web.github.io/webxr-samples/ work again.

Fixes https://github.com/exokitxr/exokit/issues/747.


modulesio:
Initial working:

Capture 27

src/XR.js
```diff
@@ -356,7 +326,7 @@ class XRWebGLLayer {
     this.alpha = alpha;
     this.multiview = multiview;
```
chrisplatorres: this can be removed per https://github.com/immersive-web/webxr/commit/bb483c993df49e94c3660e6a7b046885d1570779

modulesio: Thanks!

modulesio:
I think this should be good -- the WebXR samples are working provided we force WebXR entry. Reality tabs and other sites work the same.

- [ ] might want to add an entry button hack for immersive web samples + A-Frame, for testing
- [ ] the rest of examples need to be updated to the new API

todrobbins:
giphy

PR #1156, opened 2019-06-02
WebXR extension API docs

modulesio:
This updates the WebXR extension API documents to not be Magic Leap specific -- since we have both emulations and other hardware bindings in Exokit.

PR https://github.com/exokitxr/exokit/pull/1002 moves the API to a more generic WebXR one on `XRSession`, and this is the documentation PR for that change. They will probably go in separately.

Broken out of https://github.com/exokitxr/exokit/pull/1002.


modulesio:
Related documentation work for Planes API: https://github.com/exokitxr/exokit/issues/1245

modulesio:
We will probably be following the WebXR spec in this, so extension documentation is less important for now.

PR #1162, opened 2019-06-03
Add FPS cap for non-XR

modulesio:
This adds an FPS cap to the monitor's refresh rate in Exokit non-XR desktop mode. Mostly the point is to not burn CPU/GPU in testing.

The old behavior (uncapped) is still available under the `--uncapped`/`-c` flag for testing the fully unleashed FPS tick rate.


modulesio:
Note this affects only non-XR mode -- Exokit will still sync to the headset rate for all headsets regardless of flag.

PR #1165, opened 2019-06-03
Worker: run importScripts after setup

modulesio:
Exokit sets up the JS context in `WindowBase` on the first tick, but `Worker` was running `importScripts` on the first tick, before the initialization completed.

Therefore defer the `importScripts` call until `process.nextTick`.

### DIFF
--- a/src/Worker.js
+++ b/src/Worker.js
@@ -1,5 +1,3 @@
-(() => {
-
 const path = require('path');
 const fs = require('fs');
 const url = require('url');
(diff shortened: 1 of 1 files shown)


src/Worker.js
```diff
@@ -1,5 +1,3 @@
-(() => {
```
RangerMauve: What was the purpose of the IIFE? To prevent `var` from leaking into the global scope?

modulesio: Yeah, I think it's legacy from when we used to run the `Worker.js` bootstrap inline.

But if it's loading as a node module, which it is right now, then `node` will actually do this for us.

PR #1169, opened 2019-06-04
Reality frames: stencils+clips

modulesio:
Exokit's reality frames (3d `<iframe>`s) do not have a mechanism to limit the render volume/stencil.

However, this is a desired feature for embedders -- the parent should be able to control the rendering volume and stencil of the child frame.

This PR adds both geometry-based stencils and plane equation-based clip planes support for parent frames to impose on child frames. 

Additionally, a portal example is used to demonstrate the API usage, which is controlled via properties on the `HTMLIFrameElement` DOM element on the parent side.

lol3


modulesio:
This one can move to the debugging phase.

modulesio:
Code for generating an indented portal geometry buffer:

```
g = new THREE.BoxBufferGeometry(); g.applyMatrix(new THREE.Matrix4().makeTranslation(0, 0, -0.5)); ps = g.toNonIndexed().attributes.position.array; ps2=[]; for (let i = 0; i < ps.length/9; i++) { (ps[i*9+2] >= 0 && ps[i*9+2+3] >= 0 && ps[i*9+2+6] >= 0) || ps2.push.apply(ps2, ps.slice(i*9, (i+1)*9)); }; ps3 = Array(ps2.length); for (let i = 0; i < ps2.length/3; i++) {ps3[i*3] = ps2[i*3]; ps3[i*3+1] = ps2[i*3+2]; ps3[i*3+2] = ps2[i*3+1]; }; ps3;
```

modulesio:
Code for re-planing a projection matrix near plane to align with an arbitrary oblique user plane: https://gist.github.com/chrisplatorres/b8f348ba547f30b6414763c4687d6b44

(we need this for correct near clip on portal planes)

modulesio:
Should be working now, for both portals and stencils:

lol3

PR #1174, opened 2019-06-06
Add process to eventtarget.js

chrisplatorres:
This fixes the error `TypeError: Cannot read property 'nextTick' of undefined` when you use webrtc.

Full error:
```
TypeError: Cannot read property 'nextTick' of undefined
    at RTCPeerConnection.dispatchEvent (C:\Users\chris\github\studio\exokit\src\RTC\eventtarget.js:23:13)
    at PeerConnection.onicecandidate (C:\Users\chris\github\studio\exokit\src\RTC\peerconnection.js:84:10)
```

### DIFF
--- a/src/RTC/eventtarget.js
+++ b/src/RTC/eventtarget.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const process = require('process');
+const {process} = global;
 
 /**
  * @author mrdoob / http://mrdoob.com/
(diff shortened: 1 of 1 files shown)


src/RTC/eventtarget.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+const process = require('process');
```
modulesio: Can it be?

```
const {process} = global;
```

Would save a function call.

chrisplatorres: 1739737

chrisplatorres: that worked for me.

PR #1182, opened 2019-06-07
Added onbeforeload flag

### LINKED ISSUE #1166: Add flag to run script before window load
It'd be nice to be able to intercept the creation of new JS contexts in exokit globally.

The motivating use case for this is to be able to add guards on top of the Iframe element for custom rules on whether parent frames can access child frames' JS contexts or pointer events.

Instead of building a bunch of extra access controls, it'd be useful to handle these concerns at the level of applications being built on top of exokit.

The most flexible way to get full control over JS contexts, is to run a node module just before user content.

Ideally there should be a command line flag like `--onbeforeload` or something which specifies the path to a script to load relative to the path. Just after the global context has [loaded](https://github.com/exokitxr/exokit/blob/master/src/WindowBase.js#L378), this argument should be detected, resolved relative to the current working directory, and required.

RangerMauve:
Closes #1166 

Adds a new CLI flag, `--onbeforeload` which will invoke a Node script after each new JS context is initialized (workers/frames), but before the third party JS loads.

The script is resolved relative to `process.cwd()` so stuff like `exokit --onbeforeload ../foobar.js` is valid.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -108,6 +108,7 @@ const args = (() => {
         'xr',
         'size',
         'replace',
+        'onbeforeload'
       ],
       alias: {
         v: 'version',
@@ -149,6 +150,7 @@ const args = (() => {
       nogl: minimistArgs.nogl,
       headless: minimistArgs.headless,
       uncapped: minimistArgs.uncapped,
+      onBeforeLoad: minimistArgs.onbeforeload || false
     };
   } else {
     return {};
@@ -1044,7 +1046,7 @@ const _startTopRenderLoop = () => {
       }
     };
     _updateMeshing();
-    
+
     const _updatePlanes = () => {
       if (xrState.planeTracking[0] && !topVrPresentState.planeTracker) {
         _startFakePlaneTracker();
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -149,6 +150,7 @@ const args = (() => {
       nogl: minimistArgs.nogl,
       headless: minimistArgs.headless,
       uncapped: minimistArgs.uncapped,
+      onBeforeLoad: minimistArgs.onbeforeload || false
```
modulesio: If it's a string as specified above, `false` is probably not a good default, and not consistent with the other arguments. I think we can leave it off.

src/index.js
```diff
@@ -108,6 +108,7 @@ const args = (() => {
         'xr',
         'size',
         'replace',
+        'onbeforeload'
```
modulesio: `onbeforeload` or `onBeforeLoad`? Below it's `onBeforeLoad`.

Personally I prefer `onbeforeload`.

RangerMauve:
Cool, I've been conditioned by `standard` to use camelcase in variables, but camel case feels weird on the CLI and `onbeforeload` feels very browsery which I like.

How's this?

RangerMauve:
Woot! 🔥🔥🔥

Issue #1203, opened 2019-06-13
[Linux] GLFW 3.3 not compatible


bknill:
Node v12.2.0, Ubuntu 18.04

After getting a successful build, I can't get the app to run.

I get nothing opening, and no errors.

Here is a trace from running the basic example. 

owner@owner-Aspire-VN7-791:~/workspace/exokit/exokit$ node --inspect . https://aframe.io/a-painter/
Debugger listening on ws://127.0.0.1:9229/fd8d821d-8ba9-47c5-8bb3-01c4b1e11ad9
For help, see: https://nodejs.org/en/docs/inspector
Debugger attached.

modulesio:
Why is the inspector being run here?

modulesio:
Also, what is the VR configuration on this machine? Is there a VR device plugged in?

chrisplatorres:
Able to repro on ubuntu 18.04 without any inspector ran and on multiple node versions without any VR devices plugged in.

modulesio:
☝️ Commit for that was `9939ded`.

chrisplatorres:
https://github.com/exokitxr/exokit/issues/1203#issuecomment-501611154
was saying that it I reproduced it on master,

but later `9939ded` was a commit that tested with no regression for this specific issue. Trying a closer commit to find exactly where the regression is from.

chrisplatorres:
This does not occur with the previous native-graphics-deps version `.20` (pre glfw 3.3) on ubuntu.

modulesio:
Thanks Chris, renamed the issue accordingly.

modulesio:
From my testing, the child window hangs -- there is not even any WebGL context creation happening.

modulesio:
I managed to fix this by moving the main event loop polling to a similar model to MacOS.

bknill:
Great thanks guys love yer work!

PR #1208, opened 2019-06-13
Main thread polling for GLFW Linux

modulesio:
Exokit's GLFW integration uses a separate thread for event polling.

However MacOS didn't like that -- it only wants events handled on the main thread -- so we had a solution for moving that poll to the main thread.

It turns out that Linux with GLFW 3.3 wants things to work that way too, or else it will simply not wake the event listener on the side thread. Therefore, this PR applies the MacOS hack to the Linux case as well.

Fixes https://github.com/exokitxr/exokit/issues/1203.

### DIFF
--- a/deps/exokit-bindings/glfw/src/glfw.cc
+++ b/deps/exokit-bindings/glfw/src/glfw.cc
@@ -4,6 +4,10 @@
 
 #include <exout>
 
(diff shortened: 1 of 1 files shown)


chrisplatorres:
Tested working on Ubuntu 18.04 with 
`node . -x webvr https://aframe.io/a-painter/`

Issue #950, opened 2019-04-19
Fails to load janusweb - window.customElements.upgrade is not a function


madjin:
**Describe the bug**
 What happened, and what did you expect to happen?
 Run `node --experimental-worker . https://web.janusvr.com` in terminal and it halts at a certain point after `TypeError: window.customElements.upgrade is not a function` 
 
**To Reproduce**
Steps or commands ran to reproduce the behavior:
1. Go to 'terminal (MINGW64)'
2. Type 'node --experimental-worker . https://web.janusvr.com'
3. Scroll down to 'bottom'

**Additional context**

**System information:**
 - OS: Windows 10
 
 - Exokit Version: d38467d
 
 - Node version: v11.6.0 

 - Npm version/build log if applicable
 
 - Is your Exokit downloaded or built?
Built from source

**Screenshots**
output log: https://pastebin.com/REqVrbp6

modulesio:
Thanks, looks like we need to implement https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/upgrade.

The place to implement it: https://github.com/exokitxr/exokit/blob/d38467d6575b0d2d03c46e8f4e5df890c23238bb/src/Window.js#L147

jbaicoianu:
Hmm, JanusWeb does include a polyfill which should be setting up the window.customElements object if it's not defined by the environment.  I'm not sure if the polyfill works when running under nodejs though.

This is the polyfill library we're using, if it helps for debugging https://github.com/WebReflection/document-register-element

modulesio:
Yeah, I've run into sites using that before but it had major timing problems in A-Frame. Exokit does implement most of that polyfill natively itself -- we're just missing this particular function.

jbaicoianu:
We're not using it quite as extensively as A-Frame does - in A-Frame, all entities are custom elements, whereas we only use custom elements to define a number of custom viewer tags, eg, ```<janus-viewer>``` ```<janus-viewer-video360 src="...">```, etc.  I know they had some performance issues with keeping the DOM in sync every time the 3d engine changed attributes, but we're not using custom elements in the same way.

modulesio:
The problem seems to be that `window.customElements` is hijacked and rewritten by the site to not support. `.upgrade`, despite the fact that we nominally support it.

I'm guessing the polyfill is being confused, kicking in, and downgrading our functionality, then throwing.

modulesio:
[...]

PR #1220, opened 2019-06-18
Bugfix baseUrl derivation

modulesio:
Browsers derive worker `baseUrl` differently than for `Window`, specifically in the case where `data:` or `blob:` urls are concerned.

Exokit had several bugs here, fixed by this PR:

- Pass `baseUrl` from parent to worker
- Passthrough baseUrl for `Worker` `data:` and `blob:` source urls
- Workers baseUrl derivation should not be different from that of windows

### DIFF
--- a/src/Location.js
+++ b/src/Location.js
@@ -1,6 +1,15 @@
 const {EventEmitter} = require('events');
 const url = require('url');
-const utils = require('./utils');
+
+const {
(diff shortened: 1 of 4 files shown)


modulesio:
This is a partial fix for https://github.com/exokitxr/exokit/issues/950

PR #1234, opened 2019-06-26
For each call to `fetch`, set the User-Agent header

### LINKED ISSUE #1230: exokit DOM doesn't support google analytics
**Describe the bug**
What happened, and what did you expect to happen?

I created a basic index.html with google analytics, then deployed it using `now`. When I visit the resulting url in chrome, I see the visit in my google analytics page.

When you go to the page using exokit, the google analytics javascript executes (e.g. `ga` is defined globally, as expected). However, on the google analytics dashboard, nothing happens. (It should show that a user visited the site.)

The root problem is that exokit's `fetch` function doesn't send `Cookie: ${document.cookie}` in the request.

You can prove this by making an html page with google analytics, copying the `collect?v...` request as `fetch`, then pasting the resulting fetch into exokit's CLI. Google analytics won't see anything. If you add 'headers: { 'Cookie': document.cookie }` to the fetch options, everything works fine.

One fix is for `window-fetch` to check whether `typeof document !== 'undefined'`, and if so, set `'Cookie': document.cookie` in the headers. What do you think the best way to fix this would be?

document.cookie should probably be saved to some kind of persistent storage, too (if it isn't already).

shawwn:
Closes #1230 

For each call to `fetch` in `src/DOM.js`, this PR sets the `User-Agent` header to `window.navigator.userAgent`, where `window` is defined as `this.ownerDocument.defaultView`.

This seems to be the correct way to get the window associated with the dom element in a multi-window environment, and `window.navigator.userAgent` seems to be the correct value to pass.

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -1643,7 +1643,7 @@ class HTMLLinkElement extends HTMLLoadableElement {
     const url = _mapUrl(this.href, this.ownerDocument.defaultView);
     
     return this.ownerDocument.resources.addResource((onprogress, cb) => {
-      this.ownerDocument.defaultView.fetch(url)
+      this.ownerDocument.defaultView.fetch(url, {headers: {'user-agent': `User-Agent: ${this.ownerDocument.defaultView.navigator.userAgent}`}})
         .then(res => {
           if (res.status >= 200 && res.status < 300) {
             return res.text();
@@ -1811,7 +1811,7 @@ class HTMLScriptElement extends HTMLLoadableElement {
     const url = _mapUrl(this.src, this.ownerDocument.defaultView);
     
     return this.ownerDocument.resources.addResource((onprogress, cb) => {
-      this.ownerDocument.defaultView.fetch(url)
+      this.ownerDocument.defaultView.fetch(url, {headers: {'user-agent': `User-Agent: ${this.ownerDocument.defaultView.navigator.userAgent}`}})
         .then(res => {
           if (res.status >= 200 && res.status < 300) {
             return res.text();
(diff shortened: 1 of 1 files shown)


modulesio:
Per https://github.com/modulesio/window-fetch/pull/2, I don't think adding manual options everywhere is the right resolution.

shawwn:
Suppose the intercept feature is done.

At that point, you'd still need to decide what `User-Agent` value to pass to fetch. And that decision would presumably be made somewhere other than inside the DOM element, just before the call to `fetch`.

In a multi-window multi-worker environment, it's not possible to know what the value of `User-Agent` should be for any arbitrary DOM element. The correct `User-Agent` value is `window.navigator.userAgent`. However, given DOM element X and DOM element Y, then X and Y might have completely different values for `this.ownerDocument` (and hence, different values for their corresponding `window.navigator.userAgent`).

Suppose `User-Agent` is a static unchanging value, knowable at startup time. Then in that case, simply setting something like

```
const _fetch = fetch;
fetch = (url, opts, ...args) => {
  if (!opts.headers) { opts.headers = new Headers(); }
  if (!opts.headers['user-agent']) {
   opts.headers['user-agent'] = `User-Agent: ${ExokitUserAgent}`;
  }
  return _fetch(url, opts, ...args);
}
```
at the top of `src/DOM.js` would give you the equivalent functionality as the intercept API.

modulesio:
The purpose of interception is to generalize the above case into a central control point.

That includes:
- Different `fetch` callpoints
- `XMLHttpRequest`
- User `fetch`
- Various Workers

RangerMauve:
Yeah, I agree with @modulesio. This is a really great idea and we should be setting the header, but it'd be better to do it inside the fetch implementation rather than in the code invoking fetch.

shawwn:
Closing in favor of #1237

PR #1237, opened 2019-06-26
Set User-Agent header to window.navigator.userAgent

### LINKED ISSUE #1230: exokit DOM doesn't support google analytics

shawwn:
Closes #1230

See https://github.com/github/fetch/issues/614#issuecomment-377263549

> Browsers will typically disallow you from changing the User-Agent string. It's a security restriction. There is no way around it; you'll have to use a different header name, e.g. X-Custom-user-agent, and configure the server to read it.

This PR forces the User-Agent header of any `window.fetch` call to be `window.navigator.userAgent`. 

(This behavior seems identical to what Chrome enforces.)


modulesio:
I think this is fine, but is there a plan to cover the rest of the cases [here](https://github.com/exokitxr/exokit/pull/1234#issuecomment-505892183)?

It seems that currently this PR adds disparity by fixing the one call point only.

shawwn:
Done.

shawwn:
I updated this PR with the latest changes from master.

It seems to cover all the cases you outlined. If you feel it's sufficient, then it's ready to merge.

src/Window.js
```diff
@@ -102,6 +102,7 @@ const {
 GlobalContext.id = id;
 GlobalContext.args = args;
 GlobalContext.version = version;
+GlobalContext.userAgent = `Mozilla/5.0 (OS) AppleWebKit/999.0 (KHTML, like Gecko) Chrome/999.0.0.0 Safari/999.0 Exokit/${GlobalContext.version}`;
```
modulesio: We shouldn't add anything new to `GlobalContext` unless there is no other way, Ideally, `GlobalContext` is deleted entirely.

src/fetch.js
```diff
@@ -0,0 +1,13 @@
+const GlobalContext = require('./GlobalContext');
+const fetch = require('window-fetch');
+
+Object.assign(module.exports, fetch);
```
modulesio: This isn't the usual way we export in the rest of the codebase. It's usually

```
module.exports.fetch = fetch;
```

src/Window.js
modulesio: Can `fetch` pull from the global `navigator` instead?

src/fetch.js
shawwn: `src/fetch.js` needs to export everything that `window-fetch` exports, including `Header`, `Request`, `Response`, and `Blob`. This accomplishes that.

The idea is that `src/fetch.js` is a drop-in replacement for `window-fetch`.

src/Window.js
shawwn: Any suggestions? The only alternative I can think of is to move this line into `src/fetch.js`. But then it will still depend on `GlobalContext.version`.

src/fetch.js
modulesio: If that's the case, maybe it's best this is put upstream into `window-fetch` as an interception API?

src/Window.js
modulesio: It could be a method on `window-fetch` that sets the interception user agent.

shawwn: Okay, I've updated the code. `src/fetch.js` now pulls from `Navigator.userAgent`, which is a `static get userAgent() { .... }` property.

modulesio: Alternatively, the `navigator` should be a global in all cases, where the `userAgent` can be accessed.

modulesio: That makes sense, but I don't think it needs to be a static. We could just access the `navigator` instance without adding any new data shapes.

src/fetch.js
shawwn: Mm, I would say that overriding the `User-Agent` header and forcing it to be a specific value is a browser-specific concern. It might be nice if `window-fetch` provided a way to do that, but the tradeoff is that the API becomes more complicated.

src/Window.js
shawwn: The `navigator` instance is per-window, and there can be multiple windows. As far as I know, there isn't a concept of "the global `window` value", so `src/fetch.js` can't really call `window.navigator.userAgent` unless the user passes in `window`, which would complicate the interface.

I think exokit wanted a constant, un-configurable User-Agent string, if I remember our conversation correctly from a few months ago when I proposed a configurable User-Agent string. If so, then it seems to make sense for it to be either attached to `GlobalContext` or a static getter on `Navigator`.

src/fetch.js
modulesio: I think that's cleaner than requiring a hack to collect and re-export properties from the module.

`window-fetch` is designed to implement the needs of a "browser", and setting a User-Agent is one of the spec requirements of that.

src/Window.js
modulesio: > As far as I know, there isn't a concept of "the global window value"

There is, in the specs. For our purposes that is simply `global`, as set up by `node`. I think this should work for all of the worker cases, as well as `window`.

src/fetch.js
shawwn: It does provide a way to do that: by passing in a `{header: {['user-agent']: ...}}` value as an argument to `fetch`.

shawwn: It's not really a hack. It's the equivalent of `export * from 'window-fetch';`

modulesio: The API I'm proposing isn't the ability to pass in a header, but to remember a header for future invocations. It's similar to this closure, but without the need for any re-exporting.

shawwn: See 'aggregating modules' in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

modulesio: The functionality is fine, but it adds a very unconventional (for the current codebase) new export style, necessitated mostly because the API isn't being defined in the right place.

If this is all contained in `window-fetch` as originally proposed, `fetch.js` itself would not need to exist, nor would this be a concern future engineers need to worry about getting wrong.

shawwn: It's not sufficient to remember the header. `fetch` itself must examine the incoming options on every invocation, and specifically set `User-Agent` to the standard exokit user agent string.

Otherwise, userland javascript would be able to set their own User-Agent string by calling `fetch(url, {header: {'user-agent': ...}})`, which is a security violation.

modulesio: > It's not sufficient to remember the header. fetch itself must examine the incoming options on every invocation, and specifically set User-Agent to the standard exokit user agent string.

This can be implemented entirely in `window-fetch` as proposed.

> Otherwise, userland javascript would be able to set their own User-Agent string by calling fetch(url, {header: {'user-agent': ...}})

This is fine -- Exokit allows user code to spoof.

shawwn: It's *very* surprising as a user of `window-fetch` that *any other module* can do `fetch = require('window-fetch'); fetch.forceTheUserAgentStringToBe(x);`

That means if you add a future dependency from an untrusted author, that author can easily publish an updated package which reconfigures `window-fetch` with a malicious `User-Agent` string.

modulesio: Malicious modules are a valid concern, but a malicious module could do much worse than setting a User-Agent -- and it's not clear that that itself would be a problem regardless.

The API wouldn't necessarily need to be `forceTheUserAgentStringToBe` -- it could be options on a constructor on the import. This would be roughly equivalent to `fetch.js`, except owned by `window-fetch` instead of Exokit.

shawwn: > This is fine -- Exokit allows user code to spoof.

Chrome does not: https://developer.chrome.com/multidevice/user-agent

You can reconfigure the user-agent string via extensions (https://chrome.google.com/webstore/detail/user-agent-switcher-for-c/djflhoibgkdhkhhcedjiklpkjnoahfmg?hl=en-US) but it's switched globally; there's no such thing as overriding it on a per-request basis.

This is important to prevent userland JS from spoofing other devices. Google Analytics also relies on standard User-Agent values, so allowing users to override it can break GA.

modulesio: Those concerns are outside the scope of Exokit. Exokit is not a browser.

shawwn: Fair enough. In that case, feel free to close this PR. I'll merge it into [sweetiekit-dom](https://github.com/sweetiebird/sweetiekit-dom) which is what we bundle with [SweetieKit](https://github.com/sweetiebird/sweetiekit).

That will fix our GA issue #1230 

I was hoping to maintain parity with exokit's DOM implementation, but it sounds like it might take some time before the necessary functionality / fixes can be integrated into `window-fetch` and published. The `window-fetch` extensions also introduce risks due to the dependency on mutable global state.

It might be a better idea for us to switch over to the standard https://github.com/bitinn/node-fetch module. There are significant upstream differences, like support for `stream.pipe`. 
image

modulesio: > sounds like it might take some time before the necessary functionality / fixes can be integrated into window-fetch and published.

Which timing are does this refer to? It looks like most of that work is done in this PR.

> The window-fetch extensions also introduce risks due to the dependency on mutable global state.

What are the risks?

> There are significant upstream differences, like support for stream.pipe.

Indeed, `window-fetch` should pull from upstream.

PR #1238, opened 2019-06-26
Fix _makeChildrenProxy

shawwn:


modulesio:
What is the fix here?

modulesio:
No information was provided by the author, but from the look of this, this implements liveness for the [`HTMLCollection` API](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/children).

This would be a good start in that case. Ideally this liveness could be applied to our [existing stub](https://github.com/exokitxr/exokit/blob/88737c094c302ba377d90b6317f6b1b6ae92355c/src/DOM.js#L124).

shawwn:
there exist `.childNodes` and `.children`. One is based on the other.

This fixes `.childNodes`. Without it, any mutation to children won't result in a call to `.update()`.

Feel free to do this in a better way. I won't get around to figuring out the mutation observer API and how to use it without introducing memory leaks due to retain cycles.

PR #1244, opened 2019-06-28
Fix script tags with no src

shawwn:
This PR fixes the following cases:

```
const script = document.createElement('script');
script.innerText = "console.log('foo');";
document.body.appendChild(script);
```

```
<html>
  <head>
    <script>
      console.log('foo')
    </script>
  </head>
</html>
```

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -1739,7 +1739,7 @@ class HTMLScriptElement extends HTMLLoadableElement {
       }
     });
     this.on('attached', () => {
-      if (this.getAttribute('src') && this.isRunnable() && this.isConnected && !this.readyState) {
+      if (this.isRunnable() && this.isConnected && !this.readyState) {
         const async = this.getAttribute('async');
         _loadRun(async !== null ? async !== 'false' : true);
       }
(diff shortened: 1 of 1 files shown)


src/DOM.js
```diff
@@ -1739,7 +1739,7 @@ class HTMLScriptElement extends HTMLLoadableElement {
       }
     });
     this.on('attached', () => {
-      if (this.getAttribute('src') && this.isRunnable() && this.isConnected && !this.readyState) {
+      if (this.isRunnable() && this.isConnected && !this.readyState) {
```
modulesio: It seems for the `innerHTML` run case, that should be handled by `runNow()` instead of `_loadRun` -- there is no (down)"load".

Rather than making `loadRunNow` handle a non-loading case, can we instead check here (on `attached`) which method we need to call (the loading one or not)?

modulesio:
Was this not working before?

```
<html>
  <head>
    <script>
      console.log('foo')
    </script>
  </head>
</html>
```

I would think that would be handled by the `innerHTML` set handler.

shawwn:
Oops, yes. The innerHTML set handler does handle that case.

This PR mainly fixes this case:

```
<html>
  <head>
    <script>
      const script = document.createElement('script');
      script.innerText = 'console.log("innerHTML");'
      document.body.appendChild(script);
    </script>
  </head>
</html>
```

src/DOM.js
shawwn: I think this change is the root of the fix. It ensures that the script will always be executed.

If anything, it might be a good idea to get rid of `runNow` and always go through a single path. That single path can check whether there's any URL.

modulesio: I'm fine with the single path, but the current state of this PR confuses things by re-adding innerHTML load under a second code path.

But folding `runNow` into a single code path that later checks the presence of the url would fix the underlying concern too.

shawwn:
Closing in favor of #1246

PR #1246, opened 2019-06-28
Support <script type="module">

### LINKED ISSUE #630: Import .mjs support
With node 11 we can support `import` and `.mjs`: https://nodejs.org/api/esm.html

The relevant flag is `--experimental-modules`.

Related: https://github.com/webmixedreality/exokit/pull/601

shawwn:
Closes #630

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,9 @@
     "lint": "eslint src tests",
     "rebuild": "shx rm -rf node_modules && npm cache clean --force && npm install",
     "serve": "serve examples",
-    "start": "node .",
+    "start": "node --experimental-vm-modules .",
     "test": "node node_modules/mocha/bin/mocha --full-trace --exit --bail --timeout 5000 tests/unit/__setup.test.js tests/unit/*.test.js tests/unit/**/*.test.js",
+    "test:one": "node node_modules/mocha/bin/mocha --full-trace --exit --bail --timeout 5000 tests/unit/__setup.test.js",
     "test:watch": "npm run test -- --watch"
   },
   "repository": {
--- a/src/utils.js
+++ b/src/utils.js
@@ -6,6 +6,8 @@ const mkdirp = require('mkdirp');
 const parseIntStrict = require('parse-int');
 
 const symbols = require('./symbols');
+const vm = require('vm');
+const fetch = require('window-fetch');
 
 function _getBaseUrl(u, currentBaseUrl = '') {
   let result;
@@ -29,6 +31,9 @@ function _getBaseUrl(u, currentBaseUrl = '') {
 module.exports._getBaseUrl = _getBaseUrl;
 
 function _normalizeUrl(src, baseUrl) {
+  if (/^[/][/]/.test(src)) {
+    src = new URL(baseUrl || 'https://www.example.com').protocol + src;
+  }
   if (!/^(?:https?|data|blob):/.test(src)) {
     return new URL(src, baseUrl).href
       .replace(/^(file:\/\/)\/([a-z]:.*)$/i, '$1$2');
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -16,8 +16,9 @@
     "lint": "eslint src tests",
     "rebuild": "shx rm -rf node_modules && npm cache clean --force && npm install",
     "serve": "serve examples",
-    "start": "node .",
+    "start": "node --experimental-vm-modules .",
```
modulesio: I think this will need mobiles (Quest + Magic Leap) support.

This will probably require a `node` rebuild to `12.5.0`+. The reason we don't have this is that recent node 12 versions introduced android syscall incompatibilities. Either way, we should resolve this before merging.

src/utils.js
```diff
@@ -38,6 +43,44 @@ function _normalizeUrl(src, baseUrl) {
...
+  } else {
+    throw new Error('fetch got invalid status code: ' + res.status + ' : ' + url);
+  }
+}
+module.exports._fetch = _fetch;
+
+var _G = typeof global !== 'undefined' ? global : 
+         typeof window !== 'undefined' ? window :
+         typeof self !== 'undefined' ? self : this;
+var _context = vm.createContext(_G);
```
modulesio: Is this necessary? As far as I remember any context creation invokes a serious performance penalty to proxy across the boundary.

src/utils.js
```diff
@@ -38,6 +43,44 @@ function _normalizeUrl(src, baseUrl) {
...
+  const script = new vm.SourceTextModule(code, {context, ...opts});
+  await script.link(async (path, {url, context}) => {
+    url = _normalizeUrl(path, url);
+    const code = await fetch(url);
+    return new vm.SourceTextModule(code, {url, context});
+  });
+  script.instantiate();
+  return await script.evaluate();
+}
+module.exports._runMJS = _runMJS;
```
modulesio: I'm not sure if `_runMJS` makes sense as a util, especially since it's dealing in global DOM things (capturing the global above). I would have put this in the DOM.js where its friends are.

package.json
shawwn: hm, I'm not sure what you mean. Starting node with `--experimental-vm-modules` should work on all versions of node v11 and v12, so no extra work should be necessary.

src/utils.js
shawwn: It's the only way to match `vm.runInThisContext`. There's no `vm.getGlobalContext` function, so I had to make one myself.

We can turn it off, but then scripts won't run in `ThisContext`.

src/utils.js
shawwn: I can move it.

src/utils.js
modulesio: What's the alternative to share the context without a proxy layer?

shawwn: I don't think there is an alternative, really. Either it's running in the global context or it's not.

I might be mistaken about that.

modulesio: I'm certain there's a way, even if it requires editing `node`. Maybe it's possible to run the module in the global context without any contextification? That's what `vm.runInThisContext` does.

package.json
RangerMauve: Yeah, I've been using this with node 11 inside [webrun](http://github.com/RangerMauve/webrun)

RangerMauve:
This is awesome! Being able to use ESM with Exokit will make life so much easier.

package.json
modulesio: I didn't know this went back to Node 11. Disregard!

src/utils.js
modulesio: I looked into the node source; it looks like the context is [a wrap over `v8::Context`](https://github.com/nodejs/node/blob/8772da6ea890afbe9088b71edf98fd3023090b00/src/module_wrap.cc#L131), so we could reasonably get the "current context" into that wrap and pass it in, without any changes to `node`.

src/utils.js
```diff
@@ -29,6 +31,9 @@ function _getBaseUrl(u, currentBaseUrl = '') {
 module.exports._getBaseUrl = _getBaseUrl;
 
 function _normalizeUrl(src, baseUrl) {
+  if (/^[/][/]/.test(src)) {
+    src = new URL(baseUrl || 'https://www.example.com').protocol + src;
```
modulesio: I don't think `https:` should be assumed in the missing `baseUrl` case.

If this is an error, maybe it should throw?

src/utils.js
modulesio: Actually, looking through the node code it looks like just having no context is fine: https://github.com/nodejs/node/blob/8772da6ea890afbe9088b71edf98fd3023090b00/lib/internal/vm/source_text_module.js#L88

We would have to consider what that means for `type="module"` though.

src/utils.js
modulesio: Resolved in https://github.com/exokitxr/exokit/pull/1246.

modulesio:
Closing in favor of https://github.com/exokitxr/exokit/pull/1246.

Issue #1213, opened 2019-06-15
Support for custom protocols in resource loader


RangerMauve:
This is to follow up from our discussions on Discord and to keep stuff in one place.

It'd be nice to be able to load resources in Exokit from different protocols, such as [dat://](https://dat.foundation/).

Currently there's no single place that resources are loaded that would handle this use case. With that in mind, here's what needs to be done:

- [ ] Rework [window-xhr](https://github.com/modulesio/window-xhr/issues/5) to use fetch to consolidate resource loading
- [ ] Add intercepts to [window-fetch](https://github.com/modulesio/window-fetch) for custom protocols
- [x] Implement [dat-fetch](https://github.com/RangerMauve/dat-fetch) which taks a DatArchive instance and exposes fetch. Based on [dat-fetch](https://github.com/RangerMauve/load-dat-page/blob/master/XHRPatcher.js#L69)
- [ ] Modify the `fetch` at the [top level of exokit frames](https://github.com/exokitxr/exokit/blob/master/src/WindowBase.js#L23), and the one used for [synchronous script loading](https://github.com/exokitxr/exokit/blob/master/src/request.js#L20)  to make use of the dat-enhanced fetch. Use [node-dat-archive](https://github.com/beakerbrowser/node-dat-archive) to start.
- [ ] Test loading content from `dat://` URLs

This might also be useful for the `exokit://` protocol handler stuff that's already in there.

RangerMauve:
Currently working on the window-xhr changes. :)

Hopefully I'll be able to put in like 4-8 hours a week on this.

RangerMauve:
[window-xhr](https://github.com/modulesio/window-xhr/pull/6) has been updated to be based on fetch.

For intercepts, we can make use of [bogo-fetch](https://github.com/bergos/proto-fetch)

So the next step is to start putting this stuff together in exokit. 😁

RangerMauve:
Woot, check out the initial PR. :D

PR #1250, opened 2019-06-30
Centralize resource loading via fetch

RangerMauve:
As per #1213 this puts all the `fetch` related in one place.

Changes:
- `fetch.js` to reuse fetch logic across modules
- proto-fetch dependency for registering protocol handlers
- replaced all calls to `require('window-fetch')` to use `./fetch.js`
- Set fetch implementation for places that require `window-xhr`

Not sure what all the whitespaces changes are about. 😕 Might be a windows thing or atom getting rid of extra spaces.

### DIFF
--- a/src/fetch.js
+++ b/src/fetch.js
@@ -1,13 +1,27 @@
-const windowFetch = require('window-fetch');
-const protoFetch = require('proto-fetch')
+const windowFetch = require('window-fetch')
+const url = require('url')
+
+// This error message is what Firefox says when you fetch an unknown protocol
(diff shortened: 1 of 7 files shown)


src/fetch.js
```diff
@@ -0,0 +1,23 @@
+const windowFetch = require('window-fetch');
+const protoFetch = require('proto-fetch')
+
+const protocols = {
```
modulesio: Probably need to handle `data://` and `blob://` here as well.

RangerMauve:
Closing in favor of #1252

PR #1255, opened 2019-07-01
Support <script type="module">

modulesio:
Fixes [THREE.js examples](https://threejs.org/examples/#webgl_animation_cloth).

Contination of https://github.com/exokitxr/exokit/pull/1246.

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -21,7 +21,7 @@ const {Event, EventTarget, MessageEvent, MouseEvent, ErrorEvent} = require('./Ev
 const {_makeWindow} = require('./WindowVm');
 const GlobalContext = require('./GlobalContext');
 const symbols = require('./symbols');
-const {_elementGetter, _elementSetter, _normalizeUrl} = require('./utils');
+const {_elementGetter, _elementSetter, _normalizeUrl, _runMJS} = require('./utils');
 const {XRRigidTransform} = require('./XR');
 const {ElectronVm} = require('./electron-vm.js');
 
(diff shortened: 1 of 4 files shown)


modulesio:
Note that this adds the `node` start requirement of `--experimental-vm-modules`.

src/DOM.js
```diff
@@ -1881,13 +1920,10 @@ class HTMLScriptElement extends HTMLLoadableElement {
     });
   }
 
+>>>>>>> master
```
RangerMauve: Merge conflicts?

modulesio: Whoops.

modulesio: https://github.com/exokitxr/exokit/pull/1255/commits/74b7c1a61cc90e8ec77543cf6e130db98c6bc276

src/DOM.js
```diff
@@ -1805,62 +1805,67 @@ class HTMLScriptElement extends HTMLLoadableElement {
...
 
-          cb();
+          if (this.getAttribute('type') === 'module') {
+            opts.url = url;
+            const script = new vm.SourceTextModule(s, opts);
+            await script.link(async (url, {url: baseUrl}) => {
+              url = _mapUrl(_normalizeUrl(url, baseUrl), this.ownerDocument.defaultView);
+              const s = await _fetch(url);
+              return new vm.SourceTextModule(s, {
+                url: baseUrl,
```
shawwn: You need to use the parent module's `context` here.

```
await script.link(async (url, {url: baseUrl, context}) => {
  ...
    return new vm.SourceTextModule(s, {
      context,
```

See https://nodejs.org/api/vm.html#vm_class_vm_sourcetextmodule

modulesio: Would that be different from leaving it out and having the context be the global `window` (as intended)?

PR #1285, opened 2019-07-04
Install optional modules

modulesio:
Install `electron` optional module so CI for windows/linux picks it up.


modulesio:
We don't need this, since [`master` is passing](https://travis-ci.org/exokitxr/exokit/builds/554348991).

PR #1295, opened 2019-07-06
OpenVR SDK 1.4.18 -- Hand tracking

modulesio:
This updates the OpenVR SDK in exokit to version 1.4.18, which supports [Skeletal Input Tracking](https://github.com/ValveSoftware/openvr/wiki/SteamVR-Skeletal-Input).

Also adds  basic support for the new VRInput system, and hooks in tracking for pressure buttons and bone skeletons if the controller supports it.

Capture11

The API is that on Index, `gamepad.buttons[5..10]` are the fingers, thumb to pinky. `gamepad.bones` is a `Float32Array` containing n bones, which are a `position` (3) followed by `orientation` (4) _in local space_ by the controller.


modulesio:
This one should have hood hand tracking support in the hands example now. Only some minor logging cleanups needed.

PR #1300, opened 2019-07-09
Fix broken link to documentation in CONTRIBUTING.md

machenmusik:
--> https://exokit.org/docs/


modulesio:
How meta, thanks!

We'll move away from that docs page shortly but this works for now.

PR #1323, opened 2019-07-17
Add missing HTMLStyleElement.innerHTML getter

shawwn:
Without this PR, `document.createElement('style').innerHTML` incorrectly returns `undefined`:
```
> document.createElement('style').innerHTML
undefined
```

With this PR, `.innerHTML` is correctly returned:
```
> document.createElement('style').innerHTML
''
```


src/DOM.js
```diff
@@ -1611,6 +1611,9 @@ class HTMLStyleElement extends HTMLLoadableElement {
     this.setAttribute('type', type);
   }
 
+  get innerHTML() {
+    return super.innerHTML;
+  }
   set innerHTML(innerHTML) {
```
modulesio: Would it be more correct to also construct and attach e.g. a `Text` node in the setter?

shawwn: I'm not sure, but the intent of this PR is to fix the bug introduced by defining `set innerHTML` without defining `get innerHTML`. Forwarding to `super.innerHTML` takes care of the issue.

modulesio: Does this correctly return the set `.innerHTML`, or just default to the empty string?

If the superclass handles this, is it more appropriate to simply remove the interception on this subclass and let the parent class handle the `innerHTML` logic?

More generally, we should take the opportunity to "do the right thing" with regard to `innerHTML` setting on `HTMLStyleElement`. I'm confident many libraries in the wild depend on getting this correct.

shawwn: I think this is proof that you will go to any length not to merge pull requests from outside contributors (or at least from me). It speaks to the effectiveness of this project.

Even simple bug fixes are endlessly debated, analyzed, and random "what if?" scenarios are brought up for no apparent reason.

There is zero – and I do mean zero – reason not merge this PR. Your code for HTMLStyleElement is broken. This PR fixes it.

Merge it and make it better, if you want. Or do your usual thing of let it sit for 3 months and then do the same exact thing yourself in a slightly different way.

shawwn: (Yes, libraries do rely on HTMLStyleElement .innerHTML actually returning the innerHTML. This is exactly the problem I ran into with exokit DOM, and why I spent time submitting this PR. It actually breaks for actual websites right now.)

PR #1326, opened 2019-07-17
Fix inital value of <input> tags

shawwn:
E.g.

```
<input value="0.75">
```


src/DOM.js
```diff
@@ -712,6 +712,12 @@ class Element extends Node {
   constructor(window, tagName = 'DIV', attrs = [], value = '', location = null) {
     super(window);
 
+    for (let { name, value: v } of attrs) {
+      if (name === 'value') {
+        value = v;
```
modulesio: Would it be more appropriate to implement the functionality of `HTMLInputElement` rather than special-casing the `Element`?

shawwn: I think if `Element` has a `.value` property then `value="zzz"` in an HTML tag should always set it.

That said, according to Chrome, most elements shouldn't have a `.value` property.

modulesio: `Element` shouldn't have a `.value`, but the way things are set up right now the subclasses handle all special element logic.

We don't necessarily want to set `.value` for all elements that have that attribute. Bt we do for some special cases, including `HTMLInputElement`, which would need to be added.

Issue #1334, opened 2019-07-18
Canvas 2D uses too many contexts


modulesio:
We use Skia accelerated rendering in Exokit. This uses GL contexts, one per canvas.

Unfortunately sites out there are using many canvas 2d's -- more than some systems allow in terms of GL context creation per window.

This should be fixable by having a single GL context that covers all of the 2d canvases, with only the textures shared.

The only thing we _might_ need to be careful about is to make sure we [reset the Skia context](https://github.com/google/skia/blob/c7dced5f68c606e68a750a74b64c8af49e907398/include/gpu/GrContext.h#L88) when switching textures. However, Skia might track this itself when switching textures.

PR #1335, opened 2019-07-19
Canvas2d context reuse

modulesio:
Fixes https://github.com/exokitxr/exokit/issues/1334.

With this PR, windows now only create a single canvas2d Skia backing context for the lifetime of the window, and this context is not destroyed when any logical `CanvasRenderingContext2D` is destroyed.


modulesio:
Tested this and it works on Canvas2d-heavy sites.

PR #1333, opened 2019-07-18
Added initial dat support

RangerMauve:
Supports loading URLs from the `dat://` protocol. 

You can test it out with

```
exokit -x webvr dat://55861001e843ad2a0d0c1e1274c07cbc4fffa0a7cd83648741bc93c55c027acf/index.html
```

- Ensures that all protocols get a `://` after the protocol
- Adds dat-fetch to the protocols list
- Adds dat to the list f protocols to normalize

TODO:

- Figure out how to get the Dat libraries to work without process.nextTick
- Have the regex which detects protocols pull from the list of protocols in fetch.js?
- Configure persistent storage for Dat archives

Closes #1213

PR #1381, opened 2019-08-06
Dat support

### LINKED ISSUE #1213: Support for custom protocols in resource loader

modulesio:
Supports loading URLs from the `dat://` protocol. 

You can test it out with

```
exokit -x webvr dat://55861001e843ad2a0d0c1e1274c07cbc4fffa0a7cd83648741bc93c55c027acf/index.html
```

- Ensures that all protocols get a `://` after the protocol
- Adds dat-fetch to the protocols list
- Adds dat to the list f protocols to normalize

TODO:

- Figure out how to get the Dat libraries to work without process.nextTick
- Have the regex which detects protocols pull from the list of protocols in fetch.js?
- Configure persistent storage for Dat archives

Closes #1213

Continuation of https://github.com/exokitxr/exokit/pull/1333.


src/fetch.js
```diff
@@ -4,8 +4,13 @@ const utils = require('./utils');
 const windowFetch = require('window-fetch');
 const {Response} = windowFetch;
 const GlobalContext = require('./GlobalContext');
+const DatArchive = require('./DatArchive.bundle.js');
+console.log(DatArchive)
```
modulesio: Probably don't need this log in production.

RangerMauve:
K, I updated it to use the SDK and got the bundle compiling into `build/js/`. It's currently using the default dat storage path so that archive created with the SDK will share storage when they're loaded in Exokit and vise versa.

RangerMauve:
This might also bring exokit-web support since the SDK should "Just Work™" on the web. 🤯

RangerMauve:
Not sure if this is actually working at the moment though. Gonna test it next time I'm on my laptop. :P

modulesio:
> This might also bring exokit-web support since the SDK should Just Work™ on the web.

Cool, does it just use a gateway automatically, or something else?

RangerMauve:
Yeah! It's connecting to https://discoveryswarm.mauve.moe/

If you want I can help set up an exokit-specific one and configure it to connect there instead. (or whatever other configs we might need)

modulesio:
Yeah, would prefer not to have any third-party network dependencies. We have Exokit-specific domains we can use.

That would only matter for `exokit-web` though, right?

RangerMauve:
Yeah, it's all configurable. Also the SDK will try to auto-detect a local discovery-swarm server if it exists (which gets bundled in dat-store for instance)

modulesio:
Where would we configure that in this code?

RangerMauve:
Basically, we'll need to require `dat-sdk/promise` instead of `dat-sdk/auto` and then pass in the options into the factory. https://github.com/datproject/sdk/blob/master/index.js#L30

There's a [swarmOpts](https://github.com/datproject/sdk/blob/master/index.js#L30) object that gets passed into [discovery-swarm-web](https://github.com/RangerMauve/discovery-swarm-web) and there you can specify a `discovery` field for the server you want to use.

PR #1390, opened 2019-08-08
Add index to gamepad constructor

chrisplatorres:
This PR proposes adding  `index` to the gamepad constructor. This is to fix an issue with Babylon.js looking for this index to assign the [`browserGamepad`](https://doc.babylonjs.com/api/classes/babylon.gamepad#browsergamepad)

### DIFF
--- a/src/VR.js
+++ b/src/VR.js
@@ -126,12 +126,10 @@ class GamepadHapticActuator {
     });
   }
 }
-    var index = 0;
 class Gamepad {
-  constructor(id, hand, xrGamepad, hapticActuator) {
-    this.id = id;
+  constructor(index, id, hand, xrGamepad, hapticActuator) {
     this.index = index;
-    index++;
+    this.id = id;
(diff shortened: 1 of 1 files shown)


src/VR.js
```diff
@@ -126,9 +126,12 @@ class GamepadHapticActuator {
     });
   }
 }
+    var index = 0;
 class Gamepad {
   constructor(id, hand, xrGamepad, hapticActuator) {
     this.id = id;
+    this.index = index;
+    index++;
```
modulesio: I would probably put this as part of the constuctor parameters. I expect engines to be sensitive to the actual value in cases where e.g. the gamepads are reconstructed, and it could be expected that the indices always start at zero. even when that happens.

modulesio: Resolved in https://github.com/exokitxr/exokit/pull/1390/commits/b3054dbd89b024fba548af02380f487c1890d409

Issue #1392, opened 2019-08-08
Emukit loading throws on baseUrl


modulesio:
```
unhandled rejection: TypeError [ERR_INVALID_URL]: Invalid URL: inflate.js
    at onParseError (internal/url.js:243:9)
    at new URL (internal/url.js:319:5)
    at Object._normalizeUrl (c:\System\Users\Biatu\Dev\git\exokit\src\utils.js:36:12)
    at _normalizeUrl (c:\System\Users\Biatu\Dev\git\exokit\src\WindowBase.js:245:36)
    at getScript (c:\System\Users\Biatu\Dev\git\exokit\src\WindowBase.js:263:14)
    at importScripts (c:\System\Users\Biatu\Dev\git\exokit\src\WindowBase.js:284:32)
    at doImportScripts (evalmachine.<anonymous>:35:18)
    at evalmachine.<anonymous>:14:5
    at emit (events.js:203:13)
    at MessagePort.<anonymous> (c:\System\Users\Biatu\Dev\git\exokit\src\WindowBase.js:339:16)
```

BiatuAutMiahn:
I believe I found a typo in `src/request.js@39` `req` is supposed to be `res`?

modulesio:
Hm yes, thanks! Does that change anything?

modulesio:
https://github.com/exokitxr/exokit/commit/de46297afe9ba35318fad2ada2d687fcc6a4e6c5

BiatuAutMiahn:
well I patched `emukit/assets/frontend/bundle/js/zip.js@872` to be:
```
var DEFAULT_WORKER_SCRIPTS = {
	deflater: ['z-worker.js', 'http://127.0.0.1:8000/assets/frontend/bundle/js/deflate.js'],
	inflater: ['z-worker.js', 'http://127.0.0.1:8000/assets/frontend/bundle/js/inflate.js']
};
```

then that bug appeared.

now im resolving this error:
```
uncaught exception: ReferenceError: _elementSetter is not defined
    at FileReader.set onload [as onload] (c:\System\Users\Biatu\Dev\git\exokit\src\File.js:14:5)
    at BlobReader.readUint8Array (http://127.0.0.1:8000/assets/frontend/bundle/js/zip.js:212:18)
    at doSeek (http://127.0.0.1:8000/assets/frontend/bundle/js/zip.js:660:12)
[...]

PR #1393, opened 2019-08-09
Autodetect missing protocol in utils.js _normalizeUrl

modulesio:
Fixes https://github.com/exokitxr/exokit/issues/1392.


modulesio:
This one regresssed CryptoVoxels.

PR #1396, opened 2019-08-09
Debug _normalizeUrl()

BiatuAutMiahn:
Added some debug messaging to _normalizeUrl()


modulesio:
@BiatuAutMiahn do we still need this PR?

BiatuAutMiahn:
If the function looks good, then no

Issue #1168, opened 2019-06-03
Reality frames queryString redirect


modulesio:
Running the tabs default launch command (`-t`) does not respect redirects at the target.

```
$ node . -t https://aframe.io/a-painter
```

☝️ The above will error, even though the server will redirect to `https://aframe.io/a-painter/` and serve HTML there.

Related PR: https://github.com/exokitxr/exokit/pull/1164

PR #1408, opened 2019-09-01
Bugfix realitytab redirects

humbletim:
proposed fix for #1168 that enables redirected reality tab URLs to load correctly.

```sh
$ node . -t https://aframe.io/a-painter
THREE.WebGLRenderer 105
no xr devices
loaded root in 2D
reality tab load ok: https://aframe.io/a-painter/
A-Frame Version: 0.7.0 (Date 2018-02-05, Commit #39daf67)
...
```

### DIFF
--- a/src/DOM.js
+++ b/src/DOM.js
@@ -309,7 +309,7 @@ Node.DOCUMENT_TYPE_NODE = 10;
 Node.DOCUMENT_FRAGMENT_NODE = 11;
 module.exports.Node = Node;
 
-const _setAttributeRaw = (el, prop, value) => {
+const _setAttributeRaw = (el, prop, value, notify = true) => {
   if (prop === 'length') {
     el.attrs.length = value;
   } else {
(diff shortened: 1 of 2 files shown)


src/DOM.js
```diff
@@ -2199,7 +2199,9 @@ class HTMLIFrameElement extends HTMLSrcableElement {
                 const parentWindow = this.ownerDocument.defaultView;
                 const options = parentWindow[symbols.optionsSymbol];
 
-                url = _normalizeUrl(url, options.baseUrl);
+                url = _normalizeUrl(res.url, options.baseUrl);
```
humbletim: note: technically this line is all that's needed for resources within the redirected AFRAME url to load correctly (the other changes to do with making sure `iframe.src` has the correct URL afterwards)

src/DOM.js
```diff
@@ -2199,7 +2199,9 @@ class HTMLIFrameElement extends HTMLSrcableElement {
...
+                // passively update .src with the final url (which can differ in case of redirects)
+                _setAttributeRaw(this, 'src', url, false);
```
modulesio: Is that necessary/how `<iframe>` works in general? Or is the redirect transparent?

humbletim: probably not necessary -- want me to update to be a one liner?

note that in browsers `iframe.src` _does_ reflect the final url after redirects, but only when accessed from allowable origins; otherwise it maintains the original input url.   (at least in Firefox and Chrome)

modulesio: Interesting, thanks for the the investigation -- makes sense with the cross-origin behavior.

If it doesn't change the behavior of the fix, I'd prefer to keep the changes minimal (i.e. https://github.com/exokitxr/exokit/pull/1408#discussion-diff-319777292R2202O).

modulesio:
Thanks!

PR #1417, opened 2019-11-26
Dynamics Compressor Node

Palmer-JC:
I wanted to get the first node, Dynamics Compressor, as a template for the others.  I compiles, but has linker problems.  Probably missed a spot in the related files that need changing.  Here is console message:
```
 Creating library C:\exokit\build\Release\exokit.lib and object C:\exokit\build\Release\exokit.exp                                                                                                                                                                                                                       
AudioContext.obj : error LNK2001: unresolved external symbol "public: class v8::Local<class v8::Object> __cdecl webaudio::AudioContext::CreateDynamicsCompressor(class v8::Local<class v8::Function>,class v8::Local<class v8::Object>)" (?CreateDynamicsCompressor@AudioContext@webaudio@@QEAA?AV?$Local@VObject@v8@@@v8@@V?
$Local@VFunction@v8@@@4@V34@@Z) [C:\exokit\build\exokit.vcxproj]                                                                                                                                                                                                                                                             
C:\exokit\build\Release\exokit.node : fatal error LNK1120: 1 unresolved externals [C:\exokit\build\exokit.vcxproj]                                                                                                                                                                                                           
```

### DIFF
--- a/deps/exokit-bindings/bindings/src/bindings.cc
+++ b/deps/exokit-bindings/bindings/src/bindings.cc
@@ -115,6 +115,8 @@ Local<Object> makeAudio() {
   exports->Set(JS_STR("GainNode"), gainNodeCons);
   Local<Value> analyserNodeCons = webaudio::AnalyserNode::Initialize(isolate);
   exports->Set(JS_STR("AnalyserNode"), analyserNodeCons);
+  Local<Value> dynamicsCompressorNodeCons = webaudio::DynamicsCompressorNode::Initialize(isolate, audioParamCons);
+  exports->Set(JS_STR("DynamicsCompressorNode"), dynamicsCompressorNodeCons);
   Local<Value> pannerNodeCons = webaudio::PannerNode::Initialize(isolate, fakeAudioParamCons);
   exports->Set(JS_STR("PannerNode"), pannerNodeCons);
   Local<Value> stereoPannerNodeCons = webaudio::StereoPannerNode::Initialize(isolate, audioParamCons);
@@ -133,7 +135,7 @@ Local<Object> makeAudio() {
   exports->Set(JS_STR("MediaStreamTrack"), mediaStreamTrackCons);
   Local<Value> microphoneMediaStreamCons = webaudio::MicrophoneMediaStream::Initialize(isolate, mediaStreamTrackCons);
   exports->Set(JS_STR("MicrophoneMediaStream"), microphoneMediaStreamCons);
-  exports->Set(JS_STR("AudioContext"), webaudio::AudioContext::Initialize(isolate, audioListenerCons, audioSourceNodeCons, audioDestinationNodeCons, gainNodeCons, analyserNodeCons, pannerNodeCons, audioBufferCons, audioBufferSourceNodeCons, audioProcessingEventCons, stereoPannerNodeCons, oscillatorNodeCons, scriptProcessorNodeCons, mediaStreamTrackCons, microphoneMediaStreamCons));
+  exports->Set(JS_STR("AudioContext"), webaudio::AudioContext::Initialize(isolate, audioListenerCons, audioSourceNodeCons, audioDestinationNodeCons, gainNodeCons, analyserNodeCons, dynamicsCompressorNodeCons, pannerNodeCons, audioBufferCons, audioBufferSourceNodeCons, audioProcessingEventCons, stereoPannerNodeCons, oscillatorNodeCons, scriptProcessorNodeCons, mediaStreamTrackCons, microphoneMediaStreamCons));
 
   return scope.Escape(exports);
 }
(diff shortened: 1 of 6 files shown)


modulesio:
It looks like [this method](https://github.com/exokitxr/exokit/pull/1417/files#diff-121e7b8f57b5191f64170328bd71c737R49) was not implemented?

Palmer-JC:
got it.  Biquad filter coming before eod.

avaer:
Awesome, is this ready for review?

Palmer-JC:
Back after a few days off.  Only the Dynamics compressor is.  I left off with the Biquad Filter erroring,  see its commit for msg.

Going to add Convolver today hopefully.

Palmer-JC:
Ok, Delay node has also been added in addition to dynamics Compressor, Biquad filter, & convolver nodes.  Both Delay & convolver have yet to be tried to run.  The Biquad filter is now able to set its audio params.

Plan to still add Channel Merger & Channel Splitter nodes before beginning tests which actually generate sounds.

Palmer-JC:
Alright,  all the Nodes are there which LabSound supports, & ready for testing / review.  So far, the Dynamics Compressor & Biquad filter have been run & settings to their params performed.  I have finished the complex `BiquadFilterNode::GetFrequencyResponse()`, which takes & updates 3 Float32Arrays.  It has not been tried.

One note about the Delay node.  In webaudio, it takes a max delay time (default 1), where as LabSound takes (float sampleRate, double maxDelayTime).  I kept the first arg the max delay.  If no 2nd arg is supplied use the sample rate of the context.

Tomorrow, will start running code which exercises by using examples in mozilla.  Cannot run my stack yet, since this requires other fixes like missing context.decodeAudioData()  & playbackRate missing from buffer source.  Not sure if you want these right now too.

Palmer-JC:
Added a test file, which goes through all the Node types, does the create & sets all the values methods.  It writes the results to the console.  Here is what is produced right now on Exokit:
```
current times initial setting 0               
analyser created                              
biquad filter created                         
channel merger created                        
channel splitter created                      
constant source FAILED!                       
convolver created                             
dynamics compressor created                   
gain created                                  
oscillator created                            
panner created                                
script processor created                      
stereo panner created                         
wave shaper FAILED!                           
=====================                         
Param method Tests                            
                                              
setValueAtTime FAILED!                        
linearRampToValueAtTime FAILED!               
exponentialRampToValueAtTime FAILED!          
setTargetAtTime FAILED!                       
setValueCurveAtTime FAILED!                   
cancelScheduledValues worked                  
cancelAndHoldAtTime worked                    
current times final setting 2.0085260770975055
```
ConstantSource & Wave Shaper fail because there is no implementation.  Not a priority, but wanted the test to have them.

Delay is commented out right now as it the program to cease.

None of the `Param` methods do not do **anything**.  The cancel methods say they passed, but that is just because the method I use to test does not do anything, so it looks like it blocked it.  This definitely needs looking at.  The tests also fail in Chrome, but pass in FireFox & Edge.

There is also a short "Exokit rules!" spoken with a reverb echo using a convolver node.  It has reverb when I run it in Firefox & Edge.  Chrome gives me a message saying it was not allowed to start cause no clicks.  Maybe that is why the Params also failed.

In exokit, you hear it but there is no reverb.

Palmer-JC:
Did some more by adding a dynamics compressor into the test speech graft.  The sample still played.

More bad news however.  If you push the playback to begin into the future, it still plays immediately.  Effectively there is no queuing.

avaer:
Hm, I think LabSound does support delayed playing in its nodes...

Palmer-JC:
Their code sure has spots for it.  This sort of goes with all the `Param` methods, which essentially set the `value` member in the future, except these never seem to anything.  

I started looking at their example code, and have not come across anywhere a `Param` method was called.

avaer:
IIRC those parameter changes are processed on a change posted to the audio thread at a later point. It could be that this audio thread isn't being updated.

Palmer-JC:
I wonder about seeing what happens with an off line context?  At first, it was just a contingency, but it clearly does not work in the same way.

avaer:
Is Offline context done differently? As far as I know the main audio processing is exactly the same, the only difference being the destination node that pulls data.

Palmer-JC:
Well time would be different anyway, though now that I think about it `context.currentTime` could just be updated differently.

One I just noticed comparing `AudioContext.h` from currently in LabSound & exokit, is in the current Lab version (left), they have added methods for dispatching events manually.  I wonder if stuff has been done in this area, since you cloned?
hFileCompare

Palmer-JC:
Think I found the starting in the future problem.  There is a hardcode 0, in the when parameter of the wrapper call.  Now that I can do something about for sure, if parameters are not be looked at.

```c++
NAN_METHOD(AudioBufferSourceNode::Start) {
  // Nan::HandleScope scope;

  AudioBufferSourceNode *audioBufferSourceNode = ObjectWrap::Unwrap<AudioBufferSourceNode>(info.This());
  ((lab::FinishableSourceNode *)audioBufferSourceNode->audioNode.get())->start(0);
}
```

Palmer-JC:
Now reading the when argument & playing & stopping in the future.

PR #1422, opened 2020-04-01
Added documentation for `--xr webvr` flag

jcc10:
Added documentation for the `--xr webvr` flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,9 @@ let window = await require('./src/').load(yourUrl);
 Now you have a handle on the window object as you test your application, and
 you can set `debugger` breakpoints, inspect memory, profile CPU, etc.
 
+## Flags
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -179,6 +179,9 @@ let window = await require('./src/').load(yourUrl);
 Now you have a handle on the window object as you test your application, and
 you can set `debugger` breakpoints, inspect memory, profile CPU, etc.
 
+## Flags
+
+- `--xr webvr` Makes exokit trigger the vr entry event regardless of the webpage. If exokit is always opening in pancake mode use this to fix that.
```
avaer, line 184: This is not strictly true; it triggers WebVR mode (the dead spec before WebXR), which supports auto-entry.

Unforunately many things use that dead spec like Mozilla Hubs.

We should clarify what the flag does rather than implying it does something specifically with VR entry, which it does not.

avaer:
Thanks!

PR #1424, opened 2020-04-01
--Help flag documentation stub.

jcc10:
So that other people who actually understand what these flags do under the hood can start adding to the documentation.


avaer:
Thanks, we should probably fill out the stubs some more.

avaer:
I think for those, someone will have to walk through the code to check what they are all doing.

